PR #120, opened 2026-05-12
issue #118: remove suspend from outbound port interfaces

### LINKED ISSUE #118: `runBlocking` in Core Use Cases Bridges Suspend Provider Ports
# `runBlocking` in Core Use Cases Bridges Suspend Provider Ports

**Type:** Architectural cleanup  
**Priority:** Medium

---

## Problem

Two use-case services in the core application layer call `kotlinx.coroutines.runBlocking` to bridge suspend functions exposed by outbound provider ports:

```kotlin
// EvaluateHeatingStateService.kt
import kotlinx.coroutines.runBlocking
...
runBlocking { handleHeatingFor(area) }

// FetchSensorReadingsService.kt
import kotlinx.coroutines.runBlocking
...
val readings = runBlocking { driver.fetchReadings() }
```

`runBlocking` is a blunt instrument: it blocks the calling thread while waiting for a coroutine to complete, and its presence in domain use cases signals that the async contract of the outbound ports has leaked into the core layer. The core should not need to know whether a port implementation is synchronous or asynchronous.

## Root Cause

The outbound provider port interfaces expose `suspend` functions (driven by Ktor's async HTTP client). The use-case services, which are synchronous, compensate by wrapping each call in `runBlocking`. This is an abstraction leak: the domain is paying for an infrastructure choice (Ktor's coroutine model).

## Proposed Solutions

Two options, each with different trade-offs:

**Option A — Remove `suspend` from the outbound port interfaces (preferred).**  
Port interfaces belong to the core and should reflect domain intent, not implementation details. Make the interfaces synchronous; let the Ktor adapters handle the async-to-sync bridging internally (e.g., via `runBlocking` inside the adapter, where it is an implementation detail). The core use cases become clean and `runBlocking`-free.

**Option B — Make the use cases `suspend` throughout.**  
Propagate coroutines from the ports upward through the use cases and inbound adapters. Requires making the inbound port interfaces `suspend` as well, and ensuring the driving side (schedulers, controllers) uses a coroutine scope. More invasive but sets up a fully async pipeline.

Option A is the lower-risk change and keeps the async model as an infrastructure concern.

## Files Involved

- `src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/EvaluateHeatingStateService.kt`
- `src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/FetchSensorReadingsService.kt`
- Outbound provider port interfaces used by the above services (to remove `suspend`)
- Corresponding Ktor adapter implementations (to absorb the `runBlocking` call)

AGRFesta:


AGRFesta:
@gemini-code-assist  Thanks for the review. These suggestions would revert the architectural fix that this PR intentionally implements, so I'm declining them.
 
  A few clarifications:
 
On Sensor, OnOffActuator, MonitoredClimateArea: These are outbound port interfaces — defined in the core, implemented by infrastructure adapters. The referenced rule about "inbound port interfaces maintaining a synchronous contract" applies to the driving side (e.g. EvaluateHeatingStateUseCase, FetchSensorReadingsUseCase), not the driven side. Making outbound ports suspend forces the core domain to be aware of the infrastructure's async model — that is the abstraction leak this PR fixes.
On EvaluateHeatingStateService: Putting runBlocking back here is the exact problem described in issue #118. The use case should be insulated from whether the adapter is sync or async.
 
 On NetatmoSmarther / SwitchBotMeter: runBlocking in an infrastructure adapter is a legitimate and localised implementation detail. The thread blocked is the scheduler's own thread, not a shared coroutine dispatcher. This is the correct place for the async-to-sync bridge when the outbound port contract  is synchronous.

PR #188, opened 2026-06-13
issue #185: add GET /devices read endpoint with provider/status/featu…

### LINKED ISSUE #185: Add a read endpoint to list/search managed devices
# Add a read endpoint to list/search managed devices

## Context

Today the only way to observe the integrated devices is `POST /devices/synchronizations`,
which is a **command**: it mutates state and returns the sync delta (`newDevices` /
`updatedDevices` / `detachedDevices`). There is no read query to simply *see* what is
registered in the home — to inspect the device list you currently have to trigger a
synchronization.

The data is already fully persisted; `DevicesJdbcRepository.getAll()` exists but is not
exposed through any inbound port or controller.

## Goal

Introduce a plain, **lean** read endpoint that returns the managed devices, with optional
filters. This is the *collection resource*: a synthetic representation answering "what exists,
filterable, cheaply". Per-device relationships (areas, etc.) are intentionally **not** here —
they belong to the device detail endpoint (separate issue).

## Proposed contract

`GET /devices`

Optional query parameters (all combinable, AND semantics):

| Param      | Example      | Description                                   |
|------------|--------------|-----------------------------------------------|
| `provider` | `SWITCHBOT`  | Filter by provider.                           |
| `status`   | `PAIRED`     | Filter by device status (`PAIRED`/`DETACHED`).|
| `feature`  | `SENSOR`     | Filter by device feature.                     |

### Response `200 OK`

Reuse the existing `DeviceResponse` shape already used by the synchronization endpoint —
**no new contract to invent**:

```json
[
  {
    "uuid": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
    "status": "PAIRED",
    "deviceProviderId": "abc-123",
    "provider": "SWITCHBOT",
    "name": "Living Room Sensor",
    "features": ["SENSOR"]
  }
]
```

## Design notes

- **Lean by design:** the list stays a single-table query over `smart_home.device`. Device
  relationships (area assignments, future links) are deliberately excluded to keep this
  contract stable and cheap — see the device detail issue. This is the key decision: the list
  must not become the kitchen sink of every device relationship.
- **Pagination:** intentionally **out of scope**. For a smart home the device count is in the
  order of tens; return the full list. Add pagination only if a real need emerges.
- **Architecture:** standard hexagonal flow — a new inbound use case (`GetDevicesUseCase`)
  reading through the devices outbound port, which maps onto the existing
  `DevicesJdbcRepository.getAll()`. Apply filters at the query level (existing columns).
- **Security:** stays behind the default API key (not part of the public health-probe
  carve-out from #176).

## Acceptance criteria

- [ ] `GET /devices` returns all persisted devices using the existing `DeviceResponse` shape.
- [ ] `provider`, `status` and `feature` filters work, individually and combined.
- [ ] Empty result returns `200 OK` with an empty array (not an error).
- [ ] Endpoint is authenticated.
- [ ] Docs updated: `docs/api/devices.md` and the table in `docs/api/API_INDEX.md`.
- [ ] `CHANGELOG.md` entry under `## [Unreleased]` / `Added`.

## Out of scope

- Pagination / sorting.
- Per-device area assignments and any other relationship (device detail issue).
- Raw provider data (device diagnostics issue).

AGRFesta:
…re filters

### DIFF
--- a/app/src/main/kotlin/org/agrfesta/sh/api/controllers/DevicesController.kt
+++ b/app/src/main/kotlin/org/agrfesta/sh/api/controllers/DevicesController.kt
@@ -1,21 +1,42 @@
 package org.agrfesta.sh.api.controllers
 
 import arrow.core.Either
+import org.agrfesta.sh.api.core.application.ports.inbounds.GetDevicesUseCase
 import org.agrfesta.sh.api.core.application.ports.inbounds.RefreshDevicesUseCase
+import org.agrfesta.sh.api.core.domain.devices.DeviceFeature
+import org.agrfesta.sh.api.core.domain.devices.DeviceStatus
+import org.agrfesta.sh.api.core.domain.devices.Provider
 import org.agrfesta.sh.api.core.domain.devices.RefreshDevicesResult
+import org.agrfesta.sh.api.core.domain.failures.DeviceRepositoryError
 import org.agrfesta.sh.api.core.domain.failures.RefreshDevicesError
 import org.springframework.http.ResponseEntity
 import org.springframework.http.ResponseEntity.internalServerError
 import org.springframework.http.ResponseEntity.ok
+import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.PostMapping
 import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RequestParam
 import org.springframework.web.bind.annotation.RestController
 
 @RestController
 @RequestMapping("/devices")
 class DevicesController(
-    private val refreshDevicesUseCase: RefreshDevicesUseCase
+    private val refreshDevicesUseCase: RefreshDevicesUseCase,
+    private val getDevicesUseCase: GetDevicesUseCase
 ) {
+    @GetMapping
+    fun getDevices(
+        @RequestParam(required = false) provider: Provider?,
+        @RequestParam(required = false) status: DeviceStatus?,
+        @RequestParam(required = false) feature: DeviceFeature?
+    ): ResponseEntity<Any> =
+        when (val result = getDevicesUseCase.execute(provider, status, feature)) {
+            is Either.Left -> when (result.value) {
+                DeviceRepositoryError -> internalServerError().body(MessageResponse("Unable to retrieve devices!"))
+            }
+            is Either.Right -> ok(result.value.map { it.toResponse() })
+        }
+
     @PostMapping("/synchronizations")
     fun synchronize(): ResponseEntity<Any> =
         when (val result = refreshDevicesUseCase.execute()) {
--- a/core/src/main/kotlin/org/agrfesta/sh/api/core/application/ports/outbounds/devices/DevicesRepository.kt
+++ b/core/src/main/kotlin/org/agrfesta/sh/api/core/application/ports/outbounds/devices/DevicesRepository.kt
@@ -2,12 +2,15 @@ package org.agrfesta.sh.api.core.application.ports.outbounds.devices
 
 import arrow.core.Either
 import org.agrfesta.sh.api.core.domain.devices.Device
+import org.agrfesta.sh.api.core.domain.devices.DeviceFeature
 import org.agrfesta.sh.api.core.domain.devices.DeviceStatus
+import org.agrfesta.sh.api.core.domain.devices.Provider
 import org.agrfesta.sh.api.core.domain.devices.ProviderDeviceData
 import org.agrfesta.sh.api.core.domain.failures.DeviceCreationFailure
 import org.agrfesta.sh.api.core.domain.failures.DeviceFetchFailure
 import org.agrfesta.sh.api.core.domain.failures.DeviceRepositoryError
 import org.agrfesta.sh.api.core.domain.failures.DeviceUpdateFailure
+import org.agrfesta.sh.api.core.domain.failures.GetDevicesFailure
 import java.util.UUID
 
 /**
@@ -35,6 +38,23 @@ interface DevicesRepository {
      * or [Either.Left] with [DeviceRepositoryError] if a database error occurs.
      */
     fun getAll(): Either<DeviceRepositoryError, Collection<Device>>
+
+    /**
+     * Retrieves the persisted devices, optionally filtered.
+     *
+     * All filters are combinable with AND semantics; a `null` filter is not applied.
+     *
+     * @param provider when set, restricts the result to devices of this [Provider].
+     * @param status when set, restricts the result to devices with this [DeviceStatus].
+     * @param feature when set, restricts the result to devices exposing this [DeviceFeature].
+     * @return [Either.Right] with the matching [Device] collection (possibly empty),
+     * or [Either.Left] with [GetDevicesFailure] if a database error occurs.
+     */
+    fun getDevices(
+        provider: Provider? = null,
+        status: DeviceStatus? = null,
+        feature: DeviceFeature? = null
+    ): Either<GetDevicesFailure, Collection<Device>>
 
     /**
      * Persists a new device with the given [id].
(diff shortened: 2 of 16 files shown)


core/src/main/kotlin/org/agrfesta/sh/api/core/application/ports/outbounds/devices/DevicesRepository.kt
```diff
@@ -36,6 +39,23 @@ interface DevicesRepository {
...
+     *
+     * All filters are combinable with AND semantics; a `null` filter is not applied.
+     *
+     * @param provider when set, restricts the result to devices of this [Provider].
+     * @param status when set, restricts the result to devices with this [DeviceStatus].
+     * @param feature when set, restricts the result to devices exposing this [DeviceFeature].
+     * @return [Either.Right] with the matching [Device] collection (possibly empty),
+     * or [Either.Left] with [GetDevicesFailure] if a database error occurs.
+     */
+    fun getDevices(
```
AGRFesta, line 53: 🤖 **Reviewer** — _severity: low · confidence: medium_

Design (non-blocking): `getDevices()` now overlaps with the pre-existing `getAll()` on this same port. Called with all-null filters it returns the identical dataset, and since this PR makes `DeviceRepositoryError` implement `GetDevicesFailure`, the two methods' `Left` types are effectively isomorphic (only `DeviceRepositoryError` is reachable for either). The port therefore exposes two parallel "fetch all devices" entry points. Refactoring the four existing `getAll()` callers (Refresh/SnapshotSensorHistory/FetchSensorReadings/EvaluateHeatingState) is reasonably out of scope for this issue, but it would be worth a follow-up to consolidate `getAll()` as `getDevices()` with no filters, so the port has a single canonical read path. Flagging so the duplication is a conscious decision rather than drift.

persistence/src/main/kotlin/org/agrfesta/sh/api/persistence/jdbc/repositories/DevicesJdbcRepository.kt
```diff
@@ -34,6 +35,29 @@ class DevicesJdbcRepository(
     fun getAll(): Collection<DeviceEntity> = jdbcTemplate
         .query("""SELECT * FROM smart_home.device;""", DeviceRowMapper)
 
+    fun findDevices(
+        provider: Provider? = null,
```
AGRFesta: 🤖 **Reviewer** — _severity: low · confidence: high_

Nit: the `= null` defaults on `findDevices(provider, status, feature)` are dead weight at this layer — the only caller, `DevicesJdbcAdapter.getDevices`, always forwards all three arguments explicitly. The optional-filter contract is already expressed (with KDoc) on the inbound use case and the outbound port; this internal repository method can drop the defaults to avoid implying a calling convention that isn't used. Trivial.

AGRFesta:
🤖 **Reviewer** — review summary (round 1, sha `bc4074e736801a13993e87cbdcc9dde5c00882b2`)

GET /devices read endpoint. Clean hexagonal slice: GetDevicesUseCase (execute()) -> dedicated GetDevicesService -> DevicesRepository.getDevices() -> JDBC adapter, mirroring the existing GetAreas* naming. Sealed GetDevicesFailure, @Service only Spring annotation in :core, controller is a thin translator matching AreasController.getAll. Full test pyramid present (unit service, MVC slice with auth tests, real-Postgres adapter slice incl. AND-combination decoys, integration). Changelog + API docs (devices.md, API_INDEX.md) + Bruno collection all updated. No correctness issues found; two low-severity design/nit notes below about overlap with the existing getAll() path.

persistence/src/main/kotlin/org/agrfesta/sh/api/persistence/jdbc/repositories/DevicesJdbcRepository.kt
AGRFesta: 🤖 **Author** — _decision: accepted_

Done — dropped the `= null` defaults. They were dead weight: the only caller (`DevicesJdbcAdapter.getDevices`) always forwards all three args, and the optional-filter contract already lives on the inbound use case and the outbound port.

app/src/main/kotlin/org/agrfesta/sh/api/controllers/DevicesController.kt
```diff
@@ -1,21 +1,39 @@
...
+    @GetMapping
+    fun getDevices(
+        @RequestParam(required = false) provider: Provider?,
+        @RequestParam(required = false) status: DeviceStatus?,
+        @RequestParam(required = false) feature: DeviceFeature?
+    ): ResponseEntity<Any> =
+        when (val result = getDevicesUseCase.execute(provider, status, feature)) {
+            is Either.Left -> internalServerError().body(MessageResponse("Unable to retrieve devices!"))
+            is Either.Right -> ok(result.value.map { it.toResponse() })
+        }
```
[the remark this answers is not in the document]
AGRFesta: 🤖 **Author** — _decision: accepted_

Done — switched to a nested `when` on `result.value` handling `DeviceRepositoryError` explicitly (with an import for the failure type). This matches the `synchronize` endpoint in the same controller and gives compiler-enforced exhaustiveness over the sealed `GetDevicesFailure`.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/ports/outbounds/devices/DevicesRepository.kt
AGRFesta: 🤖 **Author** — _decision: accepted_

Agreed, and a conscious decision rather than drift. Keeping `getDevices()` and `getAll()` separate in this PR so it stays a focused read-endpoint addition without touching the four stable `getAll()` callers (Refresh / SnapshotSensorHistory / FetchSensorReadings / EvaluateHeatingState). Tracking a follow-up to consolidate `getAll()` into `getDevices()` with no filters as the single canonical read path.

persistence/src/main/kotlin/org/agrfesta/sh/api/persistence/jdbc/repositories/DevicesJdbcRepository.kt
AGRFesta: 🤖 **Reviewer**

Verified in the delta: the `= null` defaults on `findDevices(provider, status, feature)` were dropped, so the internal repository method no longer implies an optional-filter calling convention its single caller doesn't use. Resolving.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/ports/outbounds/devices/DevicesRepository.kt
AGRFesta: 🤖 **Reviewer**

Agreed — keeping this PR a focused read-endpoint addition without touching the four stable `getAll()` callers is the right call, and the duplication is now a conscious decision rather than drift. Resolving on the basis of the tracked follow-up to consolidate `getAll()` into `getDevices()` with no filters as the single canonical read path.

AGRFesta:
🤖 **Reviewer** — review summary (round 2, sha `6b34216d880b0c6983c7b32825c0c30a281ea377`)

Re-review of the delta bc4074e..6b34216 (the `fixup` commit). Both round-1 reviewer threads are addressed: the dead `= null` defaults on `findDevices` were dropped, and the design overlap between `getDevices()`/`getAll()` was accepted as a conscious, in-scope decision with a tracked follow-up. The controller's nested `when` on `result.value` is exhaustive without an `else` (verified: `GetDevicesFailure` is sealed with `DeviceRepositoryError` as its only implementor), giving the compiler-enforced exhaustiveness the gemini comment asked for. No new issues in the delta.

Issue #186, opened 2026-06-13
Add a per-device detail endpoint exposing our aggregate model


AGRFesta:
# Add a per-device detail endpoint exposing our aggregate model

## Context

The device list endpoint (separate issue) is intentionally lean: it returns only the base
device fields and no relationships. But a device participates in relationships our model knows
about — most notably the **area assignments** — and there is no way to retrieve, for a single
device, the full picture of what *we* know about it.

This is the natural home for the device **aggregate**: putting relationships in the list would
break its contract every time a new link is added (areas today, heating schedules / sensor
history / actuator state tomorrow) and pile JOINs onto an endpoint that must stay cheap. A
dedicated item resource solves this once: the list stays stable, the detail grows.

## Goal

A per-device detail endpoint returning the base device fields **plus the relationships our
model holds**, designed to accommodate future links without further contract churn.

This endpoint represents **our truth** (the persisted aggregate) — the sibling of the
diagnostics endpoint, which represents the **provider's truth** (realtime raw passthrough).

## Proposed contract

`GET /devices/{uuid}`

### Response `200 OK`

```json
{
  "uuid": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
  "name": "Living Room Sensor",
  "provider": "SWITCHBOT",
  "deviceProviderId": "abc-123",
  "status": "PAIRED",
  "features": ["SENSOR"],
  "createdOn": "2026-01-10T09:00:00Z",
  "updatedOn": "2026-06-12T18:04:00Z",
  "assignments": [
    { "areaUuid": "a1b2...", "areaName": "Living Room", "role": "SENSOR" }
  ]
}
```

### Response `404 Not Found`

Returned when `uuid` does not match any persisted device.

### Area assignments are role-scoped, multi-valued, and current-only

A device is linked to an area through **two distinct relations**:

- `sensor_assignment` — **historical**: `connected_on` / `disconnected_on`. The *current*
  assignment is the row with `disconnected_on IS NULL`. (`findByDevice` already exists.)
- `actuator_assignment` — a **plain link** (no history; created/deleted). (`findByDevice`
  already exists.)

Therefore:

- `assignments` is a **list** (empty when unassigned), not a single field — the same device
  can be a sensor in one area and an actuator in another, and the schema does not constrain it
  to a single area.
- Each entry is **role-scoped** (`SENSOR` / `ACTUATOR`).
[...]

PR #190, opened 2026-06-14
issue #187: add GET /devices/{uuid}/diagnostics realtime provider pas…

### LINKED ISSUE #187: Add a realtime per-device provider diagnostics endpoint
# Add a realtime per-device provider diagnostics endpoint

## Context

The system normalizes provider data at the adapter boundary: `DevicesProvider.getAllDevices()`
returns `ProviderDeviceData(deviceProviderId, provider, name, features)` and **discards**
everything else the provider exposes (e.g. SwitchBot's `enableCloudService`, `hubDeviceId`,
`deviceType`; Netatmo exposes much more). There is no way to see the full, unfiltered truth a
provider holds about a device, including fields we don't currently model.

Calling the provider directly from an HTTP client (e.g. Bruno) is impractical: it often
requires building signed/encrypted payloads, token handling, etc. — work the backend already
does (see SwitchBot's HMAC-signed headers, Netatmo's OAuth token refresh). In the past we
exposed dedicated "provider-shaped" endpoints for this and abandoned them.

## Goal

A single, rarely-used **diagnostic** endpoint that, given one of our devices, knows which
provider to call and returns the **realtime** raw truth from that provider — unwrapped, as
plain JSON. No persistence: this is a passthrough that must report exactly what the provider
says *right now*, including failures.

This endpoint represents the **provider's truth** — the sibling of `GET /devices/{uuid}`,
which represents **our truth** (the persisted aggregate).

## Proposed contract

`GET /devices/{uuid}/diagnostics`

- Resolve the device by `uuid` → `(provider, deviceProviderId)` from the persisted entity.
- Route to the matching provider implementation via the existing `provider` discriminator.
- Return the provider's raw response.

### Response `200 OK`

`Content-Type: application/json` — the provider's raw payload, written through **verbatim**
(no re-modelling). Shape is provider-specific and intentionally not described by a fixed schema.

### Error semantics (the diagnostic value lives here — do NOT mask failures)

| Situation                                              | Status |
|--------------------------------------------------------|--------|
| `uuid` not found among persisted devices               | `404 Not Found` |
| Provider reached but returns an error / is unreachable | `502 Bad Gateway` (surface the provider message) |
| Device's provider has no diagnostics implementation    | `501 Not Implemented` |

### New, separate port — not an extension of `DevicesProvider`

By ISP, diagnostics is a distinct capability from the sync snapshot. Introduce a dedicated
outbound port; each existing provider service (`SwitchBotService`, `NetatmoService`)
implements it **in addition** to `DevicesProvider` (same Spring bean, two interfaces).

```kotlin
interface DeviceDiagnosticsProvider {
    val provider: Provider
    fun inspect(deviceProviderId: String): Either<DevicesProviderFailure, String>
}
```

Routing reuses the pattern already used for sync (`RefreshDevicesService` injects
`Set<DevicesProvider>`): inject `Set<DeviceDiagnosticsProvider>` and select
`first { it.provider == device.provider }`.

### Return type: `String` (raw JSON body), not `JsonNode`

The port lives in `core`. Returning `JsonNode` would couple the hexagonal core to Jackson.
`String` keeps the core free of any JSON library **and** is the most faithful to the
"provider's truth": it is the literal body the provider returned, with no parse → re-serialize
round trip that could reorder keys or drop fields. The controller sets
`Content-Type: application/json` and writes it through unchanged.

Note: the provider clients already hold the raw body at the boundary (e.g.
`SwitchBotDevicesClient.getDeviceStatus()` reads `bodyAsText()` before parsing), so exposing
the raw string is a natural fit.

### "Everything about THIS device", not a filtered list row

The provider implementation decides which provider endpoint(s) to hit. The realtime value
often comes from a per-device call (e.g. SwitchBot `/devices/{id}/status`, distinct from the
device list) rather than from filtering the bulk snapshot. The contract only says: *give me
everything you know about this device id, now.*

### Security

Stays behind the default API key. Raw provider payloads may contain sensitive metadata
(`hubDeviceId`, etc.) — explicitly **outside** the public health-probe carve-out from #176.

## Acceptance criteria

- [ ] New outbound port `DeviceDiagnosticsProvider` returning `Either<DevicesProviderFailure, String>`.
- [ ] `SwitchBotService` and `NetatmoService` implement it (realtime call, raw body passthrough).
- [ ] `GET /devices/{uuid}/diagnostics` resolves the device, routes by `provider`, returns the
      raw JSON with `Content-Type: application/json`.
- [ ] Error mapping: `404` (unknown uuid), `502` (provider failure), `501` (no diagnostics impl).
- [ ] No persistence introduced; no caching.
- [ ] Endpoint is authenticated and excluded from the public carve-out.
- [ ] Docs updated: `docs/api/devices.md`, `docs/api/API_INDEX.md`, and a design note in
      `docs/domain/DEVICES.md` describing the diagnostics passthrough.
- [ ] `CHANGELOG.md` entry under `## [Unreleased]` / `Added`.

## Out of scope

- Persisting or caching the raw payload (this is intentionally realtime-only).
- Modelling/normalizing the unused provider fields into the domain.
- A bulk diagnostics endpoint (single-device only).

AGRFesta:
…sthrough

### DIFF
--- a/app/src/main/kotlin/org/agrfesta/sh/api/controllers/DevicesController.kt
+++ b/app/src/main/kotlin/org/agrfesta/sh/api/controllers/DevicesController.kt
@@ -3,6 +3,7 @@ package org.agrfesta.sh.api.controllers
 import arrow.core.Either
 import org.agrfesta.sh.api.core.application.ports.inbounds.GetDeviceUseCase
 import org.agrfesta.sh.api.core.application.ports.inbounds.GetDevicesUseCase
+import org.agrfesta.sh.api.core.application.ports.inbounds.InspectDeviceUseCase
 import org.agrfesta.sh.api.core.application.ports.inbounds.RefreshDevicesUseCase
 import org.agrfesta.sh.api.core.domain.devices.DeviceFeature
 import org.agrfesta.sh.api.core.domain.devices.DeviceStatus
@@ -10,11 +11,16 @@ import org.agrfesta.sh.api.core.domain.devices.Provider
 import org.agrfesta.sh.api.core.domain.devices.RefreshDevicesResult
 import org.agrfesta.sh.api.core.domain.failures.DeviceNotFound
 import org.agrfesta.sh.api.core.domain.failures.DeviceRepositoryError
+import org.agrfesta.sh.api.core.domain.failures.DiagnosticsNotSupported
+import org.agrfesta.sh.api.core.domain.failures.DiagnosticsProviderFailure
 import org.agrfesta.sh.api.core.domain.failures.RefreshDevicesError
+import org.springframework.http.HttpStatus
+import org.springframework.http.MediaType
 import org.springframework.http.ResponseEntity
 import org.springframework.http.ResponseEntity.internalServerError
 import org.springframework.http.ResponseEntity.notFound
 import org.springframework.http.ResponseEntity.ok
+import org.springframework.http.ResponseEntity.status
 import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.PathVariable
 import org.springframework.web.bind.annotation.PostMapping
@@ -28,7 +34,8 @@ import java.util.UUID
 class DevicesController(
     private val refreshDevicesUseCase: RefreshDevicesUseCase,
     private val getDevicesUseCase: GetDevicesUseCase,
-    private val getDeviceUseCase: GetDeviceUseCase
+    private val getDeviceUseCase: GetDeviceUseCase,
+    private val inspectDeviceUseCase: InspectDeviceUseCase
 ) {
     @GetMapping
     fun getDevices(
@@ -51,6 +58,21 @@ class DevicesController(
                 is DeviceNotFound -> notFound().build()
                 DeviceRepositoryError -> internalServerError()
                     .body(MessageResponse("Unable to retrieve device '$uuid'!"))
+            }
+        }
+
+    @GetMapping("/{uuid}/diagnostics")
+    fun diagnostics(@PathVariable uuid: UUID): ResponseEntity<Any> =
+        when (val result = inspectDeviceUseCase.execute(uuid)) {
+            is Either.Right -> ok().contentType(MediaType.APPLICATION_JSON).body(result.value)
+            is Either.Left -> when (val failure = result.value) {
+                is DeviceNotFound -> notFound().build()
+                DeviceRepositoryError -> internalServerError()
+                    .body(MessageResponse("Unable to retrieve device '$uuid'!"))
+                DiagnosticsNotSupported -> status(HttpStatus.NOT_IMPLEMENTED)
+                    .body(MessageResponse("Diagnostics is not available for device '$uuid'!"))
+                is DiagnosticsProviderFailure -> status(HttpStatus.BAD_GATEWAY)
+                    .body(MessageResponse(failure.message ?: "Diagnostics provider failed for device '$uuid'!"))
             }
         }
 
--- a/providers/src/main/kotlin/org/agrfesta/sh/api/providers/netatmo/devices/NetatmoSmarther.kt
+++ b/providers/src/main/kotlin/org/agrfesta/sh/api/providers/netatmo/devices/NetatmoSmarther.kt
@@ -11,14 +11,19 @@ import org.agrfesta.sh.api.core.domain.devices.ActuatorStatus.OFF
 import org.agrfesta.sh.api.core.domain.devices.ActuatorStatus.ON
 import org.agrfesta.sh.api.core.domain.devices.ActuatorStatus.UNDEFINED
 import org.agrfesta.sh.api.core.domain.devices.Heater
+import org.agrfesta.sh.api.core.domain.devices.Inspectable
 import org.agrfesta.sh.api.core.domain.devices.Provider
 import org.agrfesta.sh.api.core.domain.devices.Sensor
 import org.agrfesta.sh.api.core.domain.devices.SensorReadings
+import org.agrfesta.sh.api.core.domain.failures.DevicesProviderError
+import org.agrfesta.sh.api.core.domain.failures.DevicesProviderFailure
+import org.agrfesta.sh.api.providers.netatmo.NETATMO_OBJECT_MAPPER
 import org.agrfesta.sh.api.providers.netatmo.NetatmoClient
 import org.agrfesta.sh.api.providers.netatmo.NetatmoClientFailure
 import org.agrfesta.sh.api.providers.netatmo.NetatmoContractBreak
 import org.agrfesta.sh.api.providers.netatmo.NetatmoHomeStatusChange
 import org.agrfesta.sh.api.providers.netatmo.NetatmoRoomStatusChange
+import org.agrfesta.sh.api.providers.netatmo.toException
 import org.agrfesta.sh.api.utils.LoggerDelegate
 import java.math.BigDecimal
 import java.time.Duration
@@ -32,9 +37,26 @@ class NetatmoSmarther(
     private val roomId: String,
     private val client: NetatmoClient,
     private val timeProvider: TimeProvider
-) : Sensor, Heater {
+) : Sensor, Heater, Inspectable {
     private val logger by LoggerDelegate()
+    private val mapper = NETATMO_OBJECT_MAPPER
     override val provider = Provider.NETATMO
+
+    // Ktor + Jackson parsing errors and room lookup must not escape the adapter as an unmapped 500
+    @Suppress("TooGenericExceptionCaught")
+    override fun inspect(): Either<DevicesProviderFailure, String> =
+        runBlocking { client.getHomeStatusRaw(homeId) }
+            .mapLeft { DevicesProviderError(it.toException()) }
+            .flatMap { body ->
+                try {
+                    mapper.readTree(body).at("/body/home/rooms")
+                        .first { it.get("id")?.asText() == roomId }
+                        .toString()
+                        .right()
+                } catch (e: Exception) {
+                    DevicesProviderError(e).left()
+                }
+            }
 
     companion object {
         internal val MIN_SET_POINT_TEMPERATURE: BigDecimal = BigDecimal("7")
--- a/providers/src/test/kotlin/org/agrfesta/sh/api/providers/switchbot/devices/SwitchBotMeterTest.kt
+++ b/providers/src/test/kotlin/org/agrfesta/sh/api/providers/switchbot/devices/SwitchBotMeterTest.kt
@@ -1,0 +1,81 @@
+package org.agrfesta.sh.api.providers.switchbot.devices
+
+import com.fasterxml.jackson.databind.ObjectMapper
+import io.kotest.assertions.arrow.core.shouldBeLeft
+import io.kotest.assertions.arrow.core.shouldBeRight
+import io.kotest.matchers.shouldBe
+import io.kotest.matchers.types.shouldBeInstanceOf
+import io.ktor.client.engine.mock.MockEngine
+import io.ktor.client.engine.mock.respond
+import io.ktor.http.HttpHeaders
+import io.ktor.http.HttpStatusCode
+import io.ktor.http.headersOf
+import io.mockk.every
+import io.mockk.mockk
+import org.agrfesta.sh.api.core.application.ports.outbounds.RandomGenerator
+import org.agrfesta.sh.api.core.application.ports.outbounds.TimeProvider
+import org.agrfesta.sh.api.core.domain.devices.Provider
+import org.agrfesta.sh.api.core.domain.failures.DevicesProviderError
+import org.agrfesta.sh.api.providers.switchbot.SwitchBotConfiguration
+import org.agrfesta.sh.api.providers.switchbot.SwitchBotDevicesClient
+import org.agrfesta.test.mothers.aRandomUniqueString
+import org.agrfesta.test.mothers.anUrl
+import org.junit.jupiter.api.Test
+import java.time.Instant
+import java.util.UUID
+
+class SwitchBotMeterTest {
+    private val config = SwitchBotConfiguration(
+        baseUrl = anUrl(),
+        token = aRandomUniqueString(),
+        secret = aRandomUniqueString()
+    )
+    private val mapper = ObjectMapper()
+    private val timeProvider: TimeProvider = mockk()
+    private val randomGenerator: RandomGenerator = mockk()
+
+    @Test fun `inspect() returns the raw device status body verbatim`() {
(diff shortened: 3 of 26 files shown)


providers/src/main/kotlin/org/agrfesta/sh/api/providers/netatmo/devices/NetatmoSmarther.kt
```diff
@@ -32,10 +36,20 @@ class NetatmoSmarther(
...
+) : Sensor, Heater, Inspectable {
     private val logger by LoggerDelegate()
+    private val mapper = NETATMO_OBJECT_MAPPER
     override val provider = Provider.NETATMO
 
+    override fun inspect(): Either<DevicesProviderFailure, String> =
+        runBlocking { client.getHomeStatusRaw(homeId) }
+            .mapLeft { DevicesProviderError(RuntimeException(it.toString())) }
+            .map { body ->
+                mapper.readTree(body).at("/body/home/rooms")
```
AGRFesta: 🤖 **Reviewer** — _severity: high · confidence: high_

This `.map { }` block runs unguarded JSON work after the provider call: `mapper.readTree(body)` can throw on a malformed body, `.at("/body/home/rooms").first { ... }` throws `NoSuchElementException` when no room matches `roomId` (e.g. the device's room is absent from the current home status), and `it.get("id").asText()` NPEs on a room without an `id`. None of these are caught, so they escape the adapter. There is no `@ControllerAdvice` in the app, so the exception becomes an unmapped Spring 500 rather than the `502 DiagnosticsProviderFailure` this endpoint is designed to return.

This breaks ARCHITECTURE.md ("DON'T let exceptions propagate beyond the Adapter layer") and the DEVICES.md design intent that diagnostics "surfaces provider failures rather than masking them" as a mapped failure. Compare `SwitchBotMeter.inspect()`, which wraps the whole call in `try/catch -> DevicesProviderError`. Suggest wrapping the parse+extract here the same way. Note the path is untested: the only tests are the happy path and a fetch-failure; there is no "room not found / malformed body" test.

app/src/main/kotlin/org/agrfesta/sh/api/controllers/DevicesController.kt
```diff
@@ -54,6 +61,21 @@ class DevicesController(
             }
         }
 
+    @GetMapping("/{uuid}/diagnostics")
+    fun diagnostics(@PathVariable uuid: UUID): ResponseEntity<Any> =
```
AGRFesta, line 65: 🤖 **Reviewer** — _severity: medium · confidence: medium_

No integration test was added for this new flow. The two sibling read endpoints introduced just before this (#185 `GET /devices`, #186 `GET /devices/{uuid}`) each have a happy-path integration test in `DevicesIntegrationTest`, and TDD_WORKFLOW.md lists Integration as "one test per flow covering the happy path". The diagnostics passthrough is the flow most worth an end-to-end test, since its value is the full-stack wiring (resolve device -> route via `ProviderDevicesFactory` -> `Inspectable` -> write the body through verbatim with `application/json`) — exactly what the slice/unit tests mock out. (The Gemini summary claims integration tests were added, but `DevicesIntegrationTest` is unchanged in this PR.)

providers/src/main/kotlin/org/agrfesta/sh/api/providers/netatmo/devices/NetatmoSmarther.kt
```diff
@@ -32,10 +36,20 @@ class NetatmoSmarther(
...
     private val timeProvider: TimeProvider
-) : Sensor, Heater {
+) : Sensor, Heater, Inspectable {
     private val logger by LoggerDelegate()
+    private val mapper = NETATMO_OBJECT_MAPPER
     override val provider = Provider.NETATMO
 
+    override fun inspect(): Either<DevicesProviderFailure, String> =
+        runBlocking { client.getHomeStatusRaw(homeId) }
+            .mapLeft { DevicesProviderError(RuntimeException(it.toString())) }
```
AGRFesta: 🤖 **Reviewer** — _severity: low · confidence: medium_

`DevicesProviderError(RuntimeException(it.toString()))` turns the `NetatmoClientFailure` into a RuntimeException whose message is the Kotlin data-class `toString()`. The controller surfaces that verbatim in the 502 body (`failure.message`), so a client sees e.g. `NetatmoAuthFailure(exception=java.lang.RuntimeException: ...)` or `NetatmoContractBreak(message=..., response=<raw body>)` rather than a clean provider message. devices.md states "the provider's message is surfaced in the body". Worth extracting a real message from the failure (and being careful not to leak the full `response`/token context into the public 502).

providers/src/main/kotlin/org/agrfesta/sh/api/providers/netatmo/devices/NetatmoSmarther.kt
```diff
@@ -32,10 +36,20 @@ class NetatmoSmarther(
...
+            .map { body ->
+                mapper.readTree(body).at("/body/home/rooms")
+                    .first { it.get("id").asText() == roomId }
+                    .toString()
```
AGRFesta: 🤖 **Reviewer** — _severity: low · confidence: medium_

The `Inspectable` KDoc and DEVICES.md emphasize returning the "literal JSON body the provider returned ... no parse -> re-serialize round trip that could reorder keys or drop fields". The Netatmo implementation does exactly a parse -> re-serialize (`mapper.readTree(body)....toString()`) on a subtree, which can reorder keys vs. what Netatmo sent. DEVICES.md does acknowledge Netatmo "extracts its own room", so this is partly by design, but the strong "verbatim / no round trip" wording on the shared `Inspectable` interface is contradicted for this driver. Consider softening the interface KDoc to scope the "verbatim" guarantee to drivers that pass the body through unchanged (SwitchBot), since Netatmo necessarily re-serializes.

providers/src/test/kotlin/org/agrfesta/sh/api/providers/switchbot/devices/SwitchBotMeterTest.kt
```diff
@@ -0,0 +1,56 @@
...
+    private val config = SwitchBotConfiguration(
+        baseUrl = anUrl(),
+        token = aRandomUniqueString(),
+        secret = aRandomUniqueString()
+    )
+    private val mapper = ObjectMapper()
+    private val timeProvider: TimeProvider = mockk()
+    private val randomGenerator: RandomGenerator = mockk()
+
+    @Test fun `inspect() returns DevicesProviderError when the provider call fails`() {
```
AGRFesta, line 60: 🤖 **Reviewer** — _severity: low · confidence: medium_

Only the failure path of `SwitchBotMeter.inspect()` is tested here; there is no happy-path test asserting it returns the provider body verbatim. The verbatim passthrough is the meaningful contract of this driver, and `NetatmoSmarther` got both happy and failure tests — worth a symmetric guard test for the SwitchBot happy path (the client-level `getDeviceStatusRaw` test does not cover the driver's `.right()` wrapping).

AGRFesta:
🤖 **Reviewer** — review summary (round 1, sha `4085a184dbe72fb393c06a67f7fd73c37589bf07`)

Review of #187 — GET /devices/{uuid}/diagnostics realtime provider passthrough. The hexagonal layering, the new Inspectable capability (ISP), the failure modelling and the controller status mapping are clean and well-documented (API docs, DEVICES.md and CHANGELOG all updated). Main concern: NetatmoSmarther.inspect() parses/extracts the room outside any try/catch, so a missing room or malformed body throws out of the adapter (there is no @ControllerAdvice) instead of mapping to a 502 — unlike SwitchBotMeter.inspect(), which guards everything. Plus a missing integration test for the new flow and a couple of lower-severity message/doc nits.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/netatmo/devices/NetatmoSmarther.kt
AGRFesta: 🤖 **Author** — _decision: accepted_


Accepted — real bug. `inspect()` now wraps the parse + room lookup in try/catch, mapping malformed body / missing room / missing `id` to `DevicesProviderError` (→ 502) instead of letting them escape as an unmapped 500. Added a "device room absent from home status" test.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/netatmo/devices/NetatmoSmarther.kt
AGRFesta: 🤖 **Author** — _decision: accepted_


Accepted. Extracted a shared `NetatmoClientFailure.toException()` that produces a clean message and deliberately avoids leaking the raw `response`/token context (`NetatmoContractBreak` surfaces only its message). Reused in both `NetatmoSmarther.inspect()` and `NetatmoService`, removing the duplicated mapping.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/netatmo/devices/NetatmoSmarther.kt
AGRFesta: 🤖 **Author** — _decision: accepted_


Accepted. Softened the `Inspectable` KDoc: the verbatim / no-round-trip guarantee is scoped to pass-through drivers (SwitchBot); scoping drivers (Netatmo, which extracts its own room) preserve all fields but may reorder keys. Behaviour unchanged — by design as noted in DEVICES.md.

app/src/main/kotlin/org/agrfesta/sh/api/controllers/DevicesController.kt
AGRFesta: 🤖 **Author** — _decision: accepted_


Accepted. Added a happy-path integration test in `DevicesIntegrationTest` exercising the full wiring (resolve device → `ProviderDevicesFactory` → `Inspectable` → verbatim body with `application/json`) via a seeded SwitchBot device. (Confirmed the Gemini summary's claim of pre-existing integration tests was inaccurate.)

providers/src/test/kotlin/org/agrfesta/sh/api/providers/switchbot/devices/SwitchBotMeterTest.kt
AGRFesta: 🤖 **Author** — _decision: accepted_


Accepted. Added a happy-path test for `SwitchBotMeter.inspect()` using the real client + MockEngine, covering the driver's `.right()` wrapping / verbatim passthrough (not covered by the client-level test).

providers/src/main/kotlin/org/agrfesta/sh/api/providers/netatmo/devices/NetatmoSmarther.kt
AGRFesta: 🤖 **Reviewer**

Verified. `inspect()` now `flatMap`s into a `try { ... }.right() } catch (e: Exception) { DevicesProviderError(e).left() }` that wraps `mapper.readTree(body)`, the `.first { ... }` room lookup, and the `id` extraction (now null-safe `get("id")?.asText()`). Malformed body, room-not-found (`NoSuchElementException`), and missing-id all map to `DevicesProviderError` -> 502 instead of escaping the adapter as an unmapped 500, satisfying ARCHITECTURE.md and the DEVICES.md intent. The new `inspect() returns DevicesProviderError when the device room is absent from home status` test covers the path. Resolved.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/netatmo/devices/NetatmoSmarther.kt
AGRFesta: 🤖 **Reviewer**

Verified. The shared `NetatmoClientFailure.toException()` builds a clean message per case and `NetatmoContractBreak` now surfaces only its `message` (not the captured `response`), avoiding the raw-body/token leak; reused in both `NetatmoSmarther.inspect()` and `NetatmoService`, removing the duplicated mapping. Resolved.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/netatmo/devices/NetatmoSmarther.kt
AGRFesta: 🤖 **Reviewer**

Verified. The `Inspectable` KDoc now scopes the verbatim / no-round-trip guarantee to pass-through drivers (SwitchBot) and documents that scoping drivers (Netatmo) preserve all fields but may reorder keys. Wording no longer contradicts the Netatmo re-serialization. Resolved.

app/src/main/kotlin/org/agrfesta/sh/api/controllers/DevicesController.kt
AGRFesta: 🤖 **Reviewer**

Verified. `GET device diagnostics returns the provider raw payload verbatim` is now in `DevicesIntegrationTest`, seeding a SWITCHBOT device and stubbing only the `switchBotDevicesClient` MockkBean (the external HTTP boundary). It drives the full real-context wiring (resolve device -> `ProviderDevicesFactory` -> `Inspectable` -> verbatim body) and asserts 200 + `application/json` + exact body. Resolved.

providers/src/test/kotlin/org/agrfesta/sh/api/providers/switchbot/devices/SwitchBotMeterTest.kt
AGRFesta: 🤖 **Reviewer**

Verified. `inspect() returns the raw device status body verbatim` now covers the SwitchBot happy path with a real `SwitchBotDevicesClient` + `MockEngine`, asserting the driver's `.right()` verbatim passthrough — symmetric with the Netatmo happy/failure coverage and not covered by the client-level test. Resolved.

AGRFesta:
🤖 **Reviewer** — review summary (round 2, sha `fe1db3a8b5e85bb7c338fdf65934882deed38ec7`)

Re-review of #187 (delta 4085a18..fe1db3a). All five round-1 findings were accepted and fixed. Verified: (high) NetatmoSmarther.inspect() now wraps parse + room lookup in try/catch, mapping malformed body / missing room / missing id to DevicesProviderError -> 502, with a covering test; NetatmoClientFailure.toException() produces clean messages without leaking raw response/token context and is reused in NetatmoService; the missing integration test now exercises the full wiring through the real Spring context against the switchBotDevicesClient MockkBean boundary; the SwitchBot driver got a verbatim happy-path test; and the Inspectable KDoc is scoped to pass-through vs scoping drivers. No new issues in the delta. All threads converge to resolved.

PR #197, opened 2026-06-14
issue #191: persist battery level in cache via BatteryPowered capability

### LINKED ISSUE #191: Persist battery level in cache
# Enabler: persist battery level in cache

## Context
`BatteryValue` already exists as a domain reading (`core/.../domain/devices/SensorReadings.kt`), but
`FetchSensorReadingsService` only handles the `ThermoHygroDataValue` branch — the battery value returned by a
driver is currently **read and discarded**. There is therefore no place where the *last known battery level* of a
device lives, which is a prerequisite for any battery-based alerting.

Current sensor readings are **not** persisted on the DB: the latest `ThermoHygroData` is kept in **Redis** via
`SensorsCurrentReadingsCacheAdapter` under the key `sensors:{provider}:{id}:thermohygro`. Battery follows the exact
same pattern.

## Decision
- **Battery is cached, not persisted on DB.** No history is needed (same rationale as thermo-hygro data).
- The cache decouples read frequency from the provider rate limit: the polling cycle refreshes the value, any
  consumer reads it from cache on demand.
- **Battery exposure is a dedicated device capability, not a sensor concern.** "Being battery-powered" is
  orthogonal to being a `Sensor`: an `Actuator` can be battery-powered too. So battery must **not** keep riding
  inside `SensorReadings` / `Sensor.fetchReadings()`. We introduce a new capability interface that **any device
  driver** can implement (sibling of `Sensor`, `Actuator`, `Inspectable`), forcing it to expose its battery level.
- **Battery level is a numeric percentage (`Int`, 0–100), SwitchBot semantics.** SwitchBot reports `battery` as an
  int percentage (`SwitchBotModel.kt`). Netatmo instead reports a categorical `battery_state`
  (`"high"`/`"medium"`/`"low"`), and today the Netatmo provider does **not** map battery at all (it is assumed to be
  meaningful only for non-battery device types we don't have yet). We therefore adopt the numeric model now and
  **defer** Netatmo battery and any normalization of the two representations to a future issue.

## Scope
- **New `BatteryPowered` capability on `DeviceDriver`** (name TBD, e.g. `BatteryPowered`), following the existing
  ISP/capability pattern (`Sensor`, `Actuator`, `Inspectable`):
  - a single method returning the current battery level, e.g.
    `fun batteryLevel(): Either<DevicesProviderFailure, Int>` (align the failure type with the existing driver
    capabilities);
  - a driver opts in by implementing it; it is **independent** of `Sensor`/`Actuator`.
- **Retire/relocate `BatteryValue`** (currently a `SensorReadings` subtype) so battery is modelled in **one place
  only** — the new capability — and not duplicated inside sensor readings.
- Collect & cache the battery level in the polling cycle for **every** device whose driver `is BatteryPowered`
  (see technical note on the `SENSOR` filter):
  - new cache key `devices:{provider}:{id}:battery` (device-level namespace, since battery is a device concern, not
    a sensor one — distinct from the existing `sensors:…:thermohygro` key);
  - a read port method to fetch the latest cached battery level for a device (returns `null` when absent).
- Make the existing battery-reporting drivers implement the new capability (SwitchBot meter). Netatmo is **out of
  scope** here (categorical `battery_state`, no battery-powered Netatmo device modelled today).
- **Set a TTL on the battery cache key** (configurable, e.g. a small multiple of the polling interval). This is an
  intentional divergence from the no-TTL `thermohygro` key: for alerting, a device offline for too long should let
  its battery value expire → become "absent" → freeze the alert (skip-on-absent in #193), rather than keep a stale
  value re-confirming the condition forever. Use the `ttl` parameter the `Cache` port already exposes.
- Surface the battery level where it is already natural to expose device state (device aggregate / dashboard or
  diagnostics — pick the one consistent with current responses; do **not** introduce a new endpoint just for this).

## Out of scope
- Any alert/notification logic (issues #192–#194).
- Persisting a battery history (explicitly excluded).
- A dedicated battery endpoint.

## Technical notes
- Reuse the existing `Cache` port (`set/get/remove`) and the DTO + Jackson serialization approach already used by
  `SensorsCurrentReadingsCacheAdapter`. A small battery DTO or a raw int payload is acceptable — stay consistent
  with the existing adapter style.
- ❗️ **The `DeviceFeature.SENSOR` filter must not gate battery collection.** `FetchSensorReadingsService` today
  iterates only `SENSOR` devices; battery is device-level, so collection must run over **all** devices (or all
  with a battery capability), not just sensors. Either broaden the iteration or collect battery in a path
  independent of the readings loop — decide based on what keeps the service cohesive. Battery readings and
  thermo-hygro readings are now fully independent concerns.
- A driver exposes the value via the new `BatteryPowered` capability (`driver is BatteryPowered`), **not** via
  `fetchReadings()`. A device may be both a `Sensor` and `BatteryPowered`; handle the two capabilities
  independently.
- Follow TDD (`docs/TDD_WORKFLOW.md`): unit test the battery-collection branch (driver is/isn't `BatteryPowered`)
  and a slice test for the cache adapter (see the outbound-adapter slice-test convention).

## Acceptance criteria
- A `BatteryPowered` capability exists on `DeviceDriver`, independent of `Sensor`/`Actuator`, and battery is no
  longer modelled inside `SensorReadings`.
- After a polling cycle, the latest battery level of any `BatteryPowered` device (sensor **or** actuator) is
  readable from cache.
- A device whose driver is not `BatteryPowered` is unaffected (no key written, no error).
- An absent/missing cache value is returned as "no value" (not an error), so downstream consumers can treat
  absence as "not evaluable".

## Dependencies
- None. Can proceed in parallel with #192.

## Definition of done
- `CHANGELOG.md` updated under `## [Unreleased]` **only if** the battery value becomes observable through a
  documented endpoint; if purely internal, no changelog entry (per project conventions).
- API docs updated if an existing endpoint response changes (`docs/api/<resource>.md` + `API_INDEX.md`).

AGRFesta:


providers/src/main/kotlin/org/agrfesta/sh/api/providers/switchbot/devices/SwitchBotMeter.kt
```diff
@@ -22,7 +23,16 @@ class SwitchBotMeter(
...
     override val deviceProviderId: String,
     private val client: SwitchBotDevicesClient
-) : Sensor, Inspectable {
+) : Sensor, Inspectable, BatteryPowered {
+
+    // Ktor ClientRequestException (not IOException) + Jackson errors require broad Exception catch
+    @Suppress("TooGenericExceptionCaught")
+    override fun batteryLevel(): Either<DevicesProviderFailure, Int> =
+        try {
+            runBlocking { client.getDeviceStatus(deviceProviderId) }.at("/body/battery").intValue().right()
```
AGRFesta: 🤖 **Reviewer** — _severity: high · confidence: high_

❗️ Efficiency regression: a `SwitchBotMeter` is both `Sensor` and `BatteryPowered`, and `FetchSensorReadingsService.execute()` now invokes `collectReadings()` *and* `collectBattery()` on it in the same cycle (FetchSensorReadingsService.kt:42-43). `fetchReadings()` (line 50) and `batteryLevel()` (this line) each call `client.getDeviceStatus(deviceProviderId)` independently — so the same `GET /devices/{id}/status` endpoint is hit **twice per meter per polling cycle**. Previously battery was extracted from the single readings payload (the now-removed `SwitchBotSensorReadings.batteryLevel`), so this doubles SwitchBot API consumption for every meter against a daily-rate-limited provider. The orthogonal-capability design is clean, but for SwitchBot both fields come from one endpoint; consider whether battery can ride the readings response (or a per-cycle status cache) rather than re-fetching.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/GetDeviceService.kt
```diff
@@ -1,18 +1,26 @@
...
-    private val deviceAggregateRepository: DeviceAggregateRepository
+    private val deviceAggregateRepository: DeviceAggregateRepository,
+    private val deviceBatteryRepository: DeviceBatteryRepository
 ) : GetDeviceUseCase {
 
     override fun execute(deviceId: UUID): Either<GetDeviceFailure, DeviceAggregate> =
         deviceAggregateRepository.findById(deviceId)
+            .map { aggregate ->
+                aggregate.copy(
+                    batteryLevel = deviceBatteryRepository.findBy(aggregate).getOrElse { null }
```
AGRFesta: 🤖 **Reviewer** — _severity: medium · confidence: medium_

`deviceBatteryRepository.findBy(aggregate).getOrElse { null }` swallows a `BatteryLookupFailure` (e.g. Redis error, or a corrupted non-int cache value → `BatteryLookupError`) silently, collapsing it to `null` with no log. ARCHITECTURE.md explicitly lists "DON'T use `Either.getOrElse` or force-unwrap results in the service layer to avoid handling errors." Graceful degradation to `null` on a lookup failure is a defensible product choice here (battery is best-effort and shouldn't fail the whole GET), but the failure should at least be logged — `FetchSensorReadingsService` logs every battery failure (lines 71, 75), whereas the read path makes a Redis outage or cache corruption completely invisible. Suggest `.onLeft { logger.error(...) }.getOrNull()` or a `fold` that logs the Left.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/switchbot/devices/SwitchBotMeter.kt
AGRFesta: 🤖 **Reviewer** — _severity: low · confidence: medium_

`.at("/body/battery").intValue()` returns `0` when the `battery` node is absent or non-numeric (Jackson `MissingNode.intValue()` == 0). A partial/malformed provider response would therefore cache `batteryLevel = 0` — indistinguishable from a genuinely critical battery — rather than surfacing a `DevicesProviderError` or `null`. There is also no 0–100 range check anywhere, despite the KDoc on `BatteryPowered.batteryLevel()` and `DeviceBatteryRepository` asserting the range and `docs/api/devices.md` documenting `(0–100)`. Consider validating presence/range (e.g. treat a missing or out-of-range node as a failure) so the documented contract is actually enforced at the boundary.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/switchbot/devices/SwitchBotMeter.kt
```diff
@@ -40,8 +50,7 @@ class SwitchBotMeter(
             val jsonNode = runBlocking { client.getDeviceStatus(deviceProviderId) }
             SwitchBotSensorReadings(
                 temperature = jsonNode.at("/body/temperature").asText().let { Temperature.of(it) },
-                humidityInt = jsonNode.at("/body/humidity").intValue(),
-                batteryLevel = jsonNode.at("/body/battery").intValue()
+                humidityInt = jsonNode.at("/body/humidity").intValue()
```
AGRFesta, line 75: 🤖 **Reviewer** — _severity: low · confidence: low_

This PR changes `fetchReadings()` behaviour (drops the `batteryLevel` field from `SwitchBotSensorReadings`) but `SwitchBotMeterTest` still has no direct test asserting `fetchReadings()` returns the temperature/humidity `ThermoHygroDataValue`. The added tests cover `batteryLevel()` and `inspect()` only. Since the method was modified here and the test class is already being touched, this is a natural moment to add a guard test for `fetchReadings()` so its contract is pinned (the `toSensorReadings()` mapping is otherwise unexercised by any unit test).

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/FetchSensorReadingsService.kt
```diff
@@ -32,25 +37,42 @@ class FetchSensorReadingsService(
...
-                val driver = factories[device.provider]?.createDevice(device) ?: return@forEach
-                if (driver is Sensor) {
-                    val readings = driver.fetchReadings()
-                        .getOrElse { failure ->
-                            logger.error("Failed to fetch readings for device ${device.uuid}: $failure")
-                            return@forEach
+        devices.forEach { device ->
+            val driver = factories[device.provider]?.createDevice(device) ?: return@forEach
+            collectReadings(device, driver)
+            collectBattery(device, driver)
```
AGRFesta, line 43: 🤖 **Reviewer** — _severity: low · confidence: low_

`FetchSensorReadingsUseCase` / `FetchSensorReadingsService` now also collects battery for *any* `BatteryPowered` driver, including non-sensor devices (actuators) that have no `SENSOR` feature — so the use-case name no longer reflects what `execute()` does. The folding into one iteration is reasonable for efficiency (single `createDevice`), but consider whether the name/contract should be broadened (e.g. a "device polling" use case) or whether battery collection warrants its own use case, per the one-interface-one-use-case guidance in ARCHITECTURE.md. Low priority / design note.

persistence/src/main/kotlin/org/agrfesta/sh/api/cache/adapters/DeviceBatteryCacheAdapter.kt
```diff
@@ -0,0 +1,54 @@
...
+                { json -> parseBatteryLevel(json) }
+            )
+
+    override fun save(device: DeviceProviderIdentity, batteryLevel: Int): Either<BatterySaveFailure, Unit> =
+        when (val response = cache.set(device.getBatteryKey(), batteryLevel.toString(), ttl.toKotlinDuration())) {
+            is CacheError -> BatterySaveError(response.exception).left()
+            CacheOkResponse -> Unit.right()
+        }
+
+    private fun parseBatteryLevel(json: String): Either<BatteryLookupFailure, Int?> =
```
AGRFesta: 🤖 **Reviewer** — _severity: low · confidence: low_

Naming nit: the parameter is called `json`, but `save()` stores a plain integer string (`batteryLevel.toString()`), not JSON — so `parseBatteryLevel` is parsing a bare int, not a JSON document. The name appears copied from `SensorsCurrentReadingsCacheAdapter.deserialize(json)` where the value genuinely is JSON. Suggest renaming to `cached` / `raw` to avoid implying a serialization format that isn't there.

AGRFesta:
🤖 **Reviewer** — review summary (round 1, sha `3ba861001b9beac865fd1d6cff9f2e9c3137cead`)

Review of #191 (persist battery level in cache via BatteryPowered capability). The hexagonal modeling is clean: BatteryPowered as an orthogonal ISP capability, a dedicated DeviceBatteryRepository outbound port with sealed failure contracts, and a cache adapter that mirrors SensorsCurrentReadingsCacheAdapter (key-derivation extension, fold/when error mapping, TTL). Docs/changelog/Bruno are all updated. Main concern is an efficiency regression: for a SwitchBotMeter the polling cycle now hits GET /devices/{id}/status twice (readings + battery) where battery previously rode the single readings payload — doubling calls against a rate-limited provider. Secondary: GetDeviceService swallows BatteryLookupFailure via getOrElse with no log (against an explicit ARCHITECTURE.md DON'T), and batteryLevel() maps a missing/non-numeric node to 0 with no range check.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/FetchSensorReadingsService.kt
AGRFesta: 👤 **Author** — decision: rejected

Valid point — `FetchSensorReadings` is now broader than its name. #191 explicitly allowed folding battery collection into the existing polling pass ("broaden the iteration … decide based on what keeps the service cohesive"), and a single `createDevice` per device/cycle is exactly why. Renaming/splitting the use case would ripple into the inbound port, the scheduler and wiring — out of scope here. Tracking as a follow-up: rename the polling use case (e.g. `PollDevicesUseCase`) once we add more per-cycle device collection.

core/src/main/kotlin/org/agrfesta/sh/api/core/domain/failures/Battery.kt
```diff
@@ -0,0 +1,7 @@
+package org.agrfesta.sh.api.core.domain.failures
+
+sealed interface BatteryLookupFailure
+data class BatteryLookupError(val exception: Exception) : BatteryLookupFailure
+
+sealed interface BatterySaveFailure
+data class BatterySaveError(val exception: Exception) : BatterySaveFailure
```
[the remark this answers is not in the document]
AGRFesta: 👤 **Author** — decision: rejected

Thanks. The repo's established convention for outbound-port failures is `data class XxxError(val exception: Exception)` — see `DevicesProviderError`, `ReadingsLookupError`, `SensorReadingsSaveError`. `Battery.kt` deliberately mirrors its sibling `SensorsCurrentReadingsCacheAdapter` / `ReadingsLookupFailure`. ARCHITECTURE.md doesn't mandate opaque `data object` repository errors; carrying the exception is how the codebase preserves the cause, and the consumer logs it (we added a log on the GET read path in the related thread). Flipping only `Battery*` to opaque objects would make it inconsistent with every other port — if we want opaque errors + adapter-level logging everywhere, that's a separate cross-cutting refactor, not in scope for #191.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/switchbot/devices/SwitchBotMeter.kt
```diff
@@ -22,7 +24,37 @@ class SwitchBotMeter(
...
+        const val MAX_BATTERY_PERCENTAGE = 100
+    }
+
+    /**
+     * The device status payload, memoized for this driver's lifetime. A driver instance lives for a
+     * single polling cycle, during which both [fetchReadings] and [batteryLevel] are invoked: caching
+     * the status lets both read from one `GET /devices/{id}/status` call instead of hitting the
+     * rate-limited provider twice.
+     */
+    private var status: JsonNode? = null
```
AGRFesta, line 40: 🤖 **Reviewer** — _severity: low · confidence: low_

Design note on the new memoization (resolves the round-1 double-fetch correctly): `status` makes `SwitchBotMeter` stateful, and its correctness now rests on an invariant the type can't enforce — "a driver instance lives for a single polling cycle." It holds for every current caller (`FetchSensorReadingsService.execute()` and `EvaluateHeatingStateService.execute()` both `createDevice` a fresh instance per call), so this is not a bug today. But a future caller that caches/reuses a driver across cycles would silently serve stale **both** readings and battery, with no invalidation hook. The KDoc documents the assumption well; consider whether it's worth pinning with a brief comment at the `createDevice` call sites too, or leaving as-is. Low priority.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/switchbot/devices/SwitchBotMeter.kt
AGRFesta: 🤖 **Reviewer**

Verified. `deviceStatus()` memoizes the status `JsonNode` for the driver's lifetime, so `fetchReadings()` and `batteryLevel()` now share a single `GET /devices/{id}/status` per cycle. Confirmed the lifetime is genuinely one cycle: `SwitchBotDevicesFactory.createDevice` news up a fresh `SwitchBotMeter` each call, and both `FetchSensorReadingsService.execute()` and `EvaluateHeatingStateService.execute()` rebuild their driver set per invocation — so the cache cannot go stale across cycles. The new `readings and battery share a single device status call per driver instance` test pins it to `calls == 1`. The doubling regression is gone.

AGRFesta: 🤖 **Reviewer**

Verified. `batteryLevel()` now guards with `battery.isInt && battery.intValue() in 0..MAX_BATTERY_PERCENTAGE` and returns `DevicesProviderError` on a missing/non-numeric/out-of-range node instead of coercing to `0`. The `batteryLevel() returns DevicesProviderError when the battery node is missing` test pins the boundary, so the documented 0–100 contract is now enforced at the provider edge.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/GetDeviceService.kt
AGRFesta: 🤖 **Reviewer**

Verified. `resolveBatteryLevel()` replaces `getOrElse { null }` with `.onLeft { logger.error(...) }.getOrNull()`, so a Redis outage or cache corruption (`BatteryLookupError`) is now logged with the device id before degrading to `null`. The KDoc captures the best-effort rationale. Addresses the ARCHITECTURE.md "don't silently swallow Either" concern while keeping battery non-essential to the GET.

persistence/src/main/kotlin/org/agrfesta/sh/api/cache/adapters/DeviceBatteryCacheAdapter.kt
AGRFesta: 🤖 **Reviewer**

Verified. The lambda param and `parseBatteryLevel` argument are renamed `json` -> `cached`, so the name no longer implies a JSON document for what is a bare integer string.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/switchbot/devices/SwitchBotMeter.kt
AGRFesta: 🤖 **Reviewer**

Verified. `fetchReadings() maps the device status to a ThermoHygroDataValue` now pins the temperature/humidity mapping (`toSensorReadings()`), so the contract changed in this PR is exercised by a direct unit test.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/FetchSensorReadingsService.kt
AGRFesta: 🤖 **Reviewer**

Conceding. The justification is reasonable: #191 explicitly sanctioned folding battery collection into the existing pass for the single-`createDevice`-per-cycle efficiency win, and a rename would ripple into the inbound port, scheduler and wiring — out of scope here. Tracking the `PollDevicesUseCase` rename as a follow-up is the right call. Resolving.

AGRFesta:
🤖 **Reviewer** — review summary (round 2, sha `26cd9194e25b23eda499b26d02e102b140eb19b3`)

Re-review of #191 (round 2, delta 3ba8610..26cd919). All five open author-action threads are addressed cleanly: the GET read path now logs the lookup failure via onLeft + getOrNull (with KDoc justifying best-effort degradation); the adapter param was renamed json->cached; the high-severity SwitchBot double-fetch is fixed by memoizing the status JsonNode for the driver's single-cycle lifetime (verified: createDevice news up a fresh SwitchBotMeter every cycle in both FetchSensorReadingsService and EvaluateHeatingStateService, so the cache never goes stale across cycles), and a test pins it to one provider call; batteryLevel() now enforces isInt + 0..100 and surfaces DevicesProviderError on a missing/out-of-range node; and a fetchReadings() guard test was added. Targeted test classes pass. Conceding the FetchSensorReadings use-case-naming thread (reasonable out-of-scope justification, follow-up tracked). One new low/low design note: the memoization's correctness now rests on an un-enforced 'driver lives one cycle' invariant.

PR #198, opened 2026-06-15
192 alert domain model and persistence

### LINKED ISSUE #192: Alert domain model and persistence
# Alert domain model and persistence

## Context
We want a generic alerting capability. The foundational concept is the **Alert**: a *stateful* domain aggregate
that is **idempotent** — re-evaluating the same condition any number of times must not create duplicate alerts. An
alert opens when a condition first becomes true, stays `OPEN` while it persists, and is `RESOLVED` when the
condition clears.

This issue introduces the **domain + persistence + read API** for alerts, with **no automatic evaluation and no
concrete rules yet** (those are #193). It can be driven/tested through the repository and a manual or test-only
trigger.

## Decision
- **Alerts live on the DB (Flyway), nothing else.** An alert is the only piece of state that *must* survive a
  restart: runtime memory and cache are excluded, because losing it would re-fire notifications for a condition
  already known and already notified.
- The alert state is the **source of truth**; notifications (issue #194) are a projection of it.

## Scope
- `Alert` aggregate with an idempotent lifecycle / state machine.
- State machine with **hysteresis** support baked into the model: transitions are driven by two thresholds
  (`trigger` and `clear`), so a value oscillating around a single threshold cannot flap the alert open/closed.
  - `(none) --condition true--> OPEN`
  - `OPEN --condition true (re-eval)--> OPEN` (idempotent: no new alert)
  - `OPEN --condition clears (past clear threshold)--> RESOLVED`
- DB table + Flyway migration (next version: `V1_1_1`).
  - Suggested columns: `uuid` (PK), `type`, `scope` (device/provider/global) + target reference (e.g. device uuid),
    `status` (`OPEN`/`RESOLVED`), `opened_at`, `resolved_at`, `last_notified_at` (nullable, reserved for #194),
    a small `details` payload (e.g. the value that tripped it). Use timestamptz (see `V1_0_7`).
  - Enforce **at most one OPEN alert** per `(type, target)` — a partial unique index/constraint — to guarantee
    idempotency at the storage layer.
- Outbound repository port + JDBC adapter (no JPA), following the existing persistence conventions.
- `GET /alerts` read endpoint returning the currently `OPEN` alerts (the "state visible in the model" the design
  calls for). Support at least filtering by status.

## Out of scope
- Automatic evaluation / hooking into the polling cycle (#193).
- Concrete rules such as `BATTERY_LOW` (#193).
- Notifications and reminder cadence (#194).
- Manual acknowledge / snooze (#195).
- Projecting alert state into other read models — home dashboard, device status (#196).

## Decision: orphan alerts on device lifecycle
- **`DETACHED` does not touch alerts** — handled in #193 (skip-on-absent freezes them; offline is the future
  `DEVICE_DETACHED` rule's concern, not battery's).
- **Hard device deletion cannot happen today**: synchronization never deletes devices (absence → `DETACHED`, see
  `docs/domain/DEVICES.md`), so a device row backing an alert always exists and there are no true orphans now.
- **Guard for the future:** *if* a device de-registration / hard-delete is ever introduced, it must
  **cascade-resolve (or delete) that device's alerts**. Capture this as a constraint (e.g. FK with the chosen
  on-delete behaviour, or an explicit cleanup in the delete use case) so the gap is closed the day deletion lands.

## Decision: `GET /alerts` is the canonical model
`GET /alerts` is the **canonical, detailed** representation of alert state and the single source of truth on the
read side. Other read models (home dashboard, device status) will **not** re-derive or duplicate alerts: they will
expose a **minimal, derived projection** computed at read time from the same `AlertRepository` (see #196). This keeps
one source of truth and avoids pushing domain knowledge (which device belongs to which area) onto the frontend.

## Technical notes
- Keep the lifecycle logic in the **core domain** (pure, unit-testable), independent from the storage adapter.
- Model `type` as an enum/sealed type designed to be extended (`BATTERY_LOW`, later `DEVICE_DETACHED`,
  `SENSOR_STALE`, ...). Do **not** hardcode battery semantics into the Alert aggregate itself.
- Follow TDD (`docs/TDD_WORKFLOW.md`): unit tests for the state machine (open/keep/resolve, hysteresis,
  idempotency), slice test for the JDBC adapter, MVC slice + integration test for `GET /alerts`.
- Hexagonal rules per `docs/ARCHITECTURE.md`; functional error handling via `Either<Failure, T>`.
- **Use the `TimeProvider` outbound port** for `opened_at` / `resolved_at` (and the reserved `last_notified_at`) —
  never `Instant.now()` — for testability, consistent with `GetHomeDashboardService`.
- **`GET /alerts` sits behind the API-key auth** (SHA-256, `docs/SECURITY.md`), like every other endpoint.
- **No `severity` field for now.** Alerts are presence/absence; severity is deferred until a second rule with a
  different criticality justifies it (the dashboard warning icon in #196 derives from presence alone).

## Acceptance criteria
- Opening a condition twice yields a single `OPEN` alert (idempotent).
- An alert only transitions to `RESOLVED` once the value crosses the `clear` threshold, not merely the `trigger`.
- Alerts survive an application restart.
- `GET /alerts` lists currently open alerts.

## Dependencies
- None for the domain/store. Can proceed in parallel with #191.

## Definition of done
- `CHANGELOG.md`: add `- Add \`GET /alerts\` ...` under `## [Unreleased]`.
- API docs: new `docs/api/alerts.md` + row in `docs/api/API_INDEX.md`.
- Domain doc: add `docs/domain/ALERTS.md` describing the lifecycle, hysteresis and idempotency decisions.

AGRFesta:


app/src/test/kotlin/org/agrfesta/sh/api/controllers/AlertsGetControllerMvcSliceTest.kt
```diff
@@ -0,0 +1,110 @@
...
+
+        // Then
+        val response: List<AlertResponse> =
+            objectMapper.readValue(responseBody, object : TypeReference<List<AlertResponse>>() {})
+        response.shouldBeEmpty()
+    }
+
+    @Test fun `getAlerts() returns 200 with the alerts in AlertResponse shape`() {
+        // Given
+        val alert = anAlert()
```
AGRFesta: 🤖 **Reviewer** — _severity: medium · confidence: medium_

`anAlert()` defaults `openedAt = Instant.now()` (sub-second precision), but `InstantSerializer` (JacksonConfiguration) writes timestamps as **epoch seconds**, dropping nanos. After the HTTP round-trip, the deserialized `response[0].openedAt` is whole-seconds, while `alert.toResponse().openedAt` still carries nanos, so `shouldContainExactly(alert.toResponse())` will mismatch (passing only on the rare exact-second boundary). The proven pattern in `DevicesGetByIdControllerMvcSliceTest:59` uses `Instant.now().truncatedTo(ChronoUnit.SECONDS)` precisely to avoid this — apply the same here, e.g. `anAlert(openedAt = Instant.now().truncatedTo(ChronoUnit.SECONDS))`. (Flaky/failing test.)

persistence/src/main/kotlin/org/agrfesta/sh/api/persistence/jdbc/adapters/AlertsJdbcAdapter.kt
```diff
@@ -0,0 +1,42 @@
...
+
+@Service
+class AlertsJdbcAdapter(
+    private val alertsRepo: AlertsJdbcRepository
+) : AlertsRepository {
+
+    private val logger by LoggerDelegate()
+
+    override fun getAlerts(status: AlertStatus?): Either<GetAlertsFailure, Collection<Alert>> = try {
+        alertsRepo.findAlerts(status).map { it.toAlert() }.right()
```
AGRFesta, line 27: 🤖 **Reviewer** — _severity: low · confidence: medium_

The `.map { it.toAlert() }` mapping runs inside a `try` that only catches `DataAccessException`. `AlertEntity.toAlert()` can throw `IllegalArgumentException` (`AlertTarget.of` -> `UUID.fromString` / `Provider.valueOf` / `AlertScope.valueOf`) or `IllegalStateException` (`requireNotNull(resolvedAt)`) on a corrupt row. Those are not `DataAccessException`, so they escape the adapter boundary as an unmapped runtime exception instead of becoming a typed `AlertRepositoryError` — contrary to ARCHITECTURE.md "catch all infrastructure exceptions inside Outbound Adapters and convert them to typed domain failures". Low impact (DB-integrity only), but worth widening the catch or mapping inside the row mapper.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/ports/outbounds/alerts/AlertsRepository.kt
```diff
@@ -0,0 +1,35 @@
...
+    /**
+     * Persists a new [alert].
+     *
+     * Storage enforces at most one OPEN alert per `(type, target)`: attempting to create a second one
+     * yields [org.agrfesta.sh.api.core.domain.failures.AlertAlreadyOpen].
+     *
+     * @return [Either.Right] with [Unit] on success,
+     * or [Either.Left] with [AlertCreationFailure] if the alert could not be created.
+     */
+    fun create(alert: Alert): Either<AlertCreationFailure, Unit>
```
AGRFesta, line 34: 🤖 **Reviewer** — _severity: low · confidence: high_

Informational / follow-up: the port exposes `create()` but no resolve/update operation, so the `AlertTransition.Resolved` produced by `evaluateAlert` (OPEN -> RESOLVED, setting `resolvedAt`) cannot be persisted yet. Consistent with the stated issue scope ("no automatic evaluation yet"), but flagging that the persistence surface is incomplete relative to the state machine for whoever wires the evaluator next.

persistence/src/main/kotlin/org/agrfesta/sh/api/persistence/jdbc/repositories/AlertsJdbcRepository.kt
```diff
@@ -0,0 +1,64 @@
...
+            "target" to alert.target.reference,
+            "status" to alert.lifecycle.status.name,
+            "openedAt" to Timestamp.from(alert.openedAt),
+            "resolvedAt" to resolvedAt?.let { Timestamp.from(it) },
+            "details" to alert.details
+        )
+        jdbcTemplate.update(sql, params)
+    }
+
+    fun deleteAll(): Int = jdbcTemplate.update("DELETE FROM smart_home.alert", emptyMap<String, Any>())
```
AGRFesta: 🤖 **Reviewer** — _severity: low · confidence: medium_

`deleteAll()` appears unused: test isolation is handled by the schema-wide TRUNCATE in `CleanSmartHomeDatabaseExtension`, and no caller references it. `DevicesJdbcRepository.deleteAll()` is used by `DevicesIntegrationTest`, but there is no equivalent caller here. Either drop it or keep only if a future test needs it (dead code).

app/src/main/kotlin/org/agrfesta/sh/api/controllers/AlertsController.kt
```diff
@@ -0,0 +1,29 @@
...
+import org.springframework.web.bind.annotation.RestController
+
+@RestController
+@RequestMapping("/alerts")
+class AlertsController(
+    private val getAlertsUseCase: GetAlertsUseCase
+) {
+
+    @GetMapping
+    fun getAlerts(@RequestParam(required = false) status: AlertStatus?): ResponseEntity<Any> =
```
AGRFesta, line 22: 🤖 **Reviewer** — _severity: low · confidence: low_

Binding `status` to `AlertStatus?` means an invalid value (e.g. `?status=FOO`) yields Spring's default `400` via `MethodArgumentTypeMismatchException`. That's a reasonable default, but it's undocumented in `docs/api/alerts.md` (which only lists OPEN/RESOLVED) and untested. Consider a brief doc note and/or a slice test pinning the 400 behaviour so it doesn't silently change.

AGRFesta:
🤖 **Reviewer** — review summary (round 1, sha `447861976922b0fa23cf77fbab848411bb2550e1`)

Alert domain model + JDBC persistence + GET /alerts. The domain modelling is excellent: illegal states are unrepresentable (sealed AlertTarget / AlertLifecycle), evaluateAlert is pure and well-tested, the partial unique index is a sound idempotency safety net, and docs/changelog are properly updated. Main issue: the GET slice test asserts an Instant round-trip with full-precision now(), but the project InstantSerializer truncates to epoch seconds, so that test will fail/flake (the device slice test truncates to seconds for exactly this reason). A few low-severity notes follow. I deliberately did NOT raise the data-object->data-class change for AlertRepositoryError: every other *RepositoryError in :core is a data object, so the current code matches the established convention.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/ports/outbounds/alerts/AlertsRepository.kt
AGRFesta: 🧑‍💻 **Author** — _decision: rejected (intentional)_


Intentional for this PR's scope (no automatic evaluation yet). `create()` is the only persistence operation a consumer needs today; the resolve/update surface for `AlertTransition.Resolved` will land together with the evaluator that actually produces it, so it's designed against a real caller rather than speculatively. Thanks for flagging.

core/src/main/kotlin/org/agrfesta/sh/api/core/domain/failures/Alerts.kt
```diff
@@ -0,0 +1,17 @@
...
+ */
+sealed interface GetAlertsFailure
+
+/**
+ * Groups all causes of a failure while creating an alert.
+ */
+sealed interface AlertCreationFailure
+
+/** Infrastructure-level failure while accessing alert persistence. */
+data object AlertRepositoryError : GetAlertsFailure, AlertCreationFailure
```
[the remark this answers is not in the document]
AGRFesta: 🧑‍💻 **Author** — _decision: rejected_


Declining this one. Every `*RepositoryError` in `:core` is a `data object` (`DeviceRepositoryError`, `AreaRepositoryError`, `PropertyRepositoryError`, …), so the current `data object` matches the established convention here — switching only `AlertRepositoryError` to a `data class` would make it the outlier. The underlying cause is also already preserved: the adapter logs it (`logger.error(…, e)`) at the boundary before returning the typed failure. Keeping as-is.

app/src/test/kotlin/org/agrfesta/sh/api/controllers/AlertsGetControllerMvcSliceTest.kt
AGRFesta: 🤖 **Reviewer**

Verified: `anAlert(openedAt = Instant.now().truncatedTo(ChronoUnit.SECONDS))` is applied at line 63, matching the proven pattern in the device slice test. The nanos-vs-epoch-seconds mismatch against `InstantSerializer` is gone, so `shouldContainExactly` is now deterministic. Resolved.

app/src/main/kotlin/org/agrfesta/sh/api/controllers/AlertsController.kt
AGRFesta: 🤖 **Reviewer**

Verified: the `400` path is now both documented (docs/api/alerts.md: the inline note plus a dedicated `Response 400 Bad Request` section) and pinned by a slice test (`getAlerts() returns 400 when the status query param is not a valid value`). The undocumented/untested gap is closed. Resolved.

persistence/src/main/kotlin/org/agrfesta/sh/api/persistence/jdbc/adapters/AlertsJdbcAdapter.kt
AGRFesta: 🤖 **Reviewer**

Verified: `getAlerts()` now catches `IllegalArgumentException` and `IllegalStateException` around `findAlerts(...).map { it.toAlert() }` and converts them to a typed `AlertRepositoryError`, with a corrupt-row guard test (DEVICE scope + `target = 'not-a-uuid'` -> `UUID.fromString` throws IAE -> `AlertRepositoryError`). The adapter boundary now respects the ARCHITECTURE.md exception-boundary rule. Resolved.

Minor (no action needed): Kotlin's `requireNotNull(resolvedAt)` in `AlertEntity.toAlert()` actually throws `IllegalArgumentException`, not `IllegalStateException` (my round-1 note mis-attributed it), and `AlertTarget.of` also throws only IAE. So the added `IllegalStateException` branch is currently unreachable/defensive — harmless, and it future-proofs the mapper if a `check(...)` is ever introduced. Fine to keep.

persistence/src/main/kotlin/org/agrfesta/sh/api/persistence/jdbc/repositories/AlertsJdbcRepository.kt
AGRFesta: 🤖 **Reviewer**

Verified: `deleteAll()` was removed from `AlertsJdbcRepository`. Confirmed no remaining reference to it for the alert table. Resolved.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/ports/outbounds/alerts/AlertsRepository.kt
AGRFesta: 🤖 **Reviewer**

Conceded. The YAGNI framing is sound: the resolve/update surface for `AlertTransition.Resolved` is best designed against the real evaluator caller rather than speculatively, and the issue scope explicitly excludes automatic evaluation. This was informational only. Resolved.

AGRFesta:
🤖 **Reviewer** — review summary (round 2, sha `bc0e045d0cd55ff4fd11aa334405edb8ab4b3bdc`)

Round 2 re-review of the Alert domain/persistence PR. All four round-1 reviewer findings are resolved in the fixup delta: the GET slice test now truncates `openedAt` to seconds (the Instant round-trip flake is gone), the `?status=FOO` 400 behaviour is now both documented in docs/api/alerts.md and pinned by a slice test, the row-mapping exception path in AlertsJdbcAdapter.getAlerts is now caught and converted to a typed AlertRepositoryError (with a corrupt-row guard test), and the unused deleteAll() was removed. The two author rejections (data-object vs data-class for AlertRepositoryError; deferring the resolve/update persistence surface) are both reasonable and conceded. No new issues in the delta. Recommending all reviewer threads resolved.

Issue #204, opened 2026-06-19
Introduce a pure heating decision (snapshot + command) in the domain


AGRFesta:
# Introduce a pure heating decision (snapshot + command) in the domain

**Parent:** #200
**Severity:** High (the architectural core of the parent issue)
**Depends on:** nothing structurally; best sequenced after #203 
**Type:** New domain logic (TDD applies)

## Summary

Extract the heating decision into a **pure domain function** that operates on an immutable snapshot
and returns a desired command — no I/O, no logging — mirroring the already-clean `evaluateAlert`
state machine. This is built **alongside** the existing strategy classes; nothing is wired in yet, so
the slice is additive and low-risk. Wiring and deletion of the old classes happen in #205.

## Target shapes (domain, `core/domain/heating/`)

```kotlin
data class HeatableAreaSnapshot(
    val areaId: UUID,
    val currentTemperature: Temperature?,   // null = reading unavailable
    val targetTemperature: Temperature?,    // null = no target configured
    val heaterStatus: ActuatorStatus        // ON / OFF / UNDEFINED (UNDEFINED when unknown)
)

enum class HeaterCommand { ON, OFF, NONE }   // NONE = nothing to do / not enough data
```

A pure entry point per strategy, e.g.:
```kotlin
fun decideComfort(areas: Collection<HeatableAreaSnapshot>): HeaterCommand
fun decideEconomy(areas: Collection<HeatableAreaSnapshot>, threshold: Percentage): HeaterCommand
```
(Exact factoring — free functions vs a `HeatingDecision` object keyed by `SharedHeatingStrategy` — to
be settled in Phase 0; the constraint is **purity**.)

## Behaviour to port (from the current strategies — preserve exactly)

- `HYSTERESIS = 1°C` (`AbstractSharedHeatingAreasStrategyService.HYSTERESIS`).
- "requires heating" for an area given its target:
  - `current > target + H` → false
  - `current < target - H` → true
  - in-band (`target - H .. target + H`) → depends on `heaterStatus`: `ON` → true, `OFF` → false,
    `UNDEFINED` → `current <= target`.
  - `currentTemperature == null` (unavailable) → treated as **not requiring** (today: logged + `false`).
  - `targetTemperature == null` → area does not require heating.
- **Comfort**: `ON` if at least one area requires heating, else `OFF`.
- **Economy**:
  - if any area is above range (`current > target + H`) → `OFF` (anti-overheat).
[...]

Issue #205, opened 2026-06-19
Rewire `EvaluateHeatingStateService` to snapshot + command; remove I/O from strategies


AGRFesta:
# Rewire `EvaluateHeatingStateService` to snapshot + command; remove I/O from strategies

**Parent:** #200
**Severity:** High
**Depends on:** 200.2 (pure decision must exist first)
**Type:** Refactor with behaviour-boundary change (TDD applies)

## Summary

Make the use case the **I/O orchestrator**: read sensor readings and heater status via ports, build
the `HeatableAreaSnapshot`s, call the pure decision from 200.2, then apply the resulting
`HeaterCommand` via ports. After this slice the heating decision no longer performs any I/O, and the
old strategy `*Service` classes are deleted.

## Scope

In:
- `EvaluateHeatingStateService` builds snapshots and applies commands.
- Replace `SharedHeatingAreasStrategyService` / `AbstractSharedHeatingAreasStrategyService` /
  `ComfortAreasSharedHeatingStrategyService` / `EconomyAreasSharedHeatingStrategyService` with the
  pure decision; keep a thin **selector** for the dynamic strategy choice
  (`DynamicSharedHeatingStrategyService` logic: read `heating.strategy` property, fall back to default).
- `HeatingConfiguration` wiring updated accordingly.

Out (later slices):
- `Area` domain still exposes the live-I/O methods at this point — they are simply no longer used by
  the heating path (purged in 200.4).
- Capability interface relocation (200.5).

## Design decision to confirm

Where does `currentTemperature` come from for the snapshot?
- **(A) Live fetch** via the sensor port (preserves today's exact behaviour:
  `MonitoredClimateArea.getCurrentTemperature()` fetches readings on each evaluation).
- **(B) Read last persisted reading** from `SensorsCurrentReadingsRepository` (what the dashboard
  uses; avoids extra provider calls but changes timing semantics).

Default recommendation: **(A)** to keep this slice behaviour-preserving; consider (B) as a separate
follow-up.

## Phase 0 — proposed test list (draft)

1. `heating.enabled` missing/false/unreadable → no device or actuator interaction.
2. Device fetch failure → evaluation skipped (no actuator commands).
3. Area fetch failure → evaluation skipped.
4. Areas grouped by shared heater → one decision per heater group.
5. Decision `ON` → `heater.on()` invoked; `OFF` → `heater.off()`; `NONE` → neither.
[...]

Issue #200, opened 2026-06-19
Device capability interfaces are outbound ports disguised as domain entities


AGRFesta:
# Device capability interfaces are outbound ports disguised as domain entities

**Severity:** Critical (architectural root cause)
**Area:** `:core` domain model, `:providers`
**Type:** Architecture / DDD / Hexagonal boundary

## Summary

The device "capability" interfaces — `Sensor`, `OnOffActuator`/`Heater`, `BatteryPowered`,
`Inspectable` — are declared in `core/domain/devices` as if they were rich domain behaviour.
Their only implementations (`NetatmoSmarther`, `SwitchBotMeter`, `SwitchBotMiniHub`) live in
`:providers`, hold an infrastructure client, and implement every method as a blocking remote I/O
call. In other words: these are **outbound ports**, and the driver classes are **adapters** — but
both are modelled and named as domain concepts.

The dependency graph still holds (`:providers` depends on `:core`), so the compile-time hexagonal
boundary is not broken. What is broken is the *purpose* of hexagonal architecture: ports are meant
to keep I/O at the edges. Here they have been smuggled into the middle of the domain model, and the
domain now performs hidden, synchronous, blocking network I/O while pretending to be pure.

This is the root cause of several other smells (domain-layer logging, the `JsonNode` memoization
band-aid, provider-specific failure types narrowing core contracts).

## Evidence

- `core/domain/devices/Sensor.kt`, `OnOffActuator.kt`, `SharedHeater.kt`, `BatteryPowered.kt`,
  `Inspectable.kt` — capability interfaces sitting in the **domain** package.
- `providers/.../switchbot/devices/SwitchBotMeter.kt`:
  - holds `private val client: SwitchBotDevicesClient` (the outbound dependency)
  - every method is `runBlocking { client... }`
  - keeps `private var status: JsonNode?` memoized "for this driver's lifetime … a driver instance
    lives for a single polling cycle" — an infrastructure-coupled lifecycle and a Jackson type held
    by a so-called domain object, existing purely to avoid double-hitting a rate-limited API.
- `providers/.../netatmo/devices/NetatmoSmarther.kt`:
  - holds `NetatmoClient`, performs `runBlocking { client.getHomeStatus(...) }` in `fetchReadings`,
    `getActuatorStatus`, `on`, `off`
  - **logs from inside the model** (`LoggerDelegate`) and narrows the core return type to the
    provider-specific `NetatmoClientFailure`.
[...]

PR #209, opened 2026-06-20
issue #204: add pure heating decision (snapshot + command) in the domain

AGRFesta:


core/src/main/kotlin/org/agrfesta/sh/api/core/domain/heating/HeatingDecision.kt
```diff
@@ -0,0 +1,58 @@
...
+ * Returns the desired [HeaterCommand] for a shared heater given the snapshots of the areas it serves
+ * and the minimum [threshold] ratio of areas that must require heating to turn the heater ON.
+ * No I/O, no logging.
+ */
+fun decideEconomy(areas: Collection<HeatableAreaSnapshot>, threshold: Percentage): HeaterCommand = when {
+    areas.isEmpty() -> HeaterCommand.NONE
+    areas.any { it.isAboveRange() } -> HeaterCommand.OFF
+    else -> {
+        val demandRatio = BigDecimal(areas.count { it.requiresHeating() })
+            .divide(BigDecimal(areas.size), 2, RoundingMode.HALF_UP)
```
AGRFesta: 🤖 **Reviewer** — _severity: medium · confidence: medium_

Rounding the demand ratio to scale 2 with `HALF_UP` before the `>=` comparison can flip the decision OFF→ON at the boundary. Example: 5 of 8 areas demand heat → true ratio 0.625, which `HALF_UP` rounds to 0.63; with `threshold = 0.63` (Percentage allows arbitrary precision in 0..1), `0.63 >= 0.63` yields ON even though the real ratio 0.625 is below the threshold. The rounding here is an unnecessary lossy step in an otherwise pure/exact decision. Prefer an exact integer/cross-multiplication comparison, e.g. `BigDecimal(count) >= threshold.value * BigDecimal(areas.size)`, so the threshold is honoured without precision artefacts.

core/src/main/kotlin/org/agrfesta/sh/api/core/domain/heating/HeatingDecision.kt
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * Returns the desired [HeaterCommand] for a shared heater given the snapshots of the areas it serves
+ * and the minimum [threshold] ratio of areas that must require heating to turn the heater ON.
+ * No I/O, no logging.
+ */
+fun decideEconomy(areas: Collection<HeatableAreaSnapshot>, threshold: Percentage): HeaterCommand = when {
+    areas.isEmpty() -> HeaterCommand.NONE
+    areas.any { it.isAboveRange() } -> HeaterCommand.OFF
+    else -> {
+        val demandRatio = BigDecimal(areas.count { it.requiresHeating() })
```
AGRFesta: 🤖 **Reviewer** — _severity: medium · confidence: low_

The denominator is `areas.size`, so areas with a missing reading (`currentTemperature == null`) or no `targetTemperature` count toward the total but can never contribute to `requiresHeating()`. They therefore dilute the demand ratio and can suppress heating when sensor data is partially unavailable. Note this is asymmetric with `decideComfort`, where such areas are simply ignored by `any { ... }`. Is diluting the ratio the intended ECONOMY semantics, or should undecidable areas be excluded from the denominator? There's no test exercising this case, which suggests it wasn't a deliberate decision — worth pinning down and covering with a test either way.

core/src/main/kotlin/org/agrfesta/sh/api/core/domain/heating/HeatingDecision.kt
```diff
@@ -0,0 +1,58 @@
+package org.agrfesta.sh.api.core.domain.heating
+
+import org.agrfesta.sh.api.core.domain.commons.Percentage
+import org.agrfesta.sh.api.core.domain.commons.Temperature
+import org.agrfesta.sh.api.core.domain.devices.ActuatorStatus
+import java.math.BigDecimal
+import java.math.RoundingMode
+
+private val HYSTERESIS: Temperature = Temperature.of(BigDecimal.ONE)
```
AGRFesta, line 9: 🤖 **Reviewer** — _severity: low · confidence: medium_

`HYSTERESIS` and the in-band/`requiresHeating()` logic here duplicate `AbstractSharedHeatingAreasStrategyService.HYSTERESIS` and `requiresHeatingFor` / `requiresHeatingInHysteresisRange` almost verbatim. This PR introduces the pure version but doesn't wire the existing strategy service to use it, so there are now two sources of truth for the same hysteresis rule that can silently drift. Acceptable as a transitional step of the heating refactor if a follow-up replaces the impure logic with these pure functions — please confirm that's the plan (and ideally reference the tracking issue) so this duplication is short-lived.

AGRFesta:
🤖 **Reviewer** — review summary (round 1, sha `92e6c5deee792e98deff29a9c0d5eab654779a89`)

Review of #204 — pure heating decision (snapshot + command) in the domain. The decision functions are clean, side-effect-free and well-tested, and correctly live in core/domain with no inbound-port machinery. Findings are about the ECONOMY ratio computation (rounding precision and treatment of areas with missing data) and the transitional duplication of the hysteresis logic with AbstractSharedHeatingAreasStrategyService.

core/src/main/kotlin/org/agrfesta/sh/api/core/domain/heating/HeatingDecision.kt
AGRFesta: Confirmed — this is a deliberate transitional step. #204 only introduces the pure decision functions; it does not wire them in. The existing `AbstractSharedHeatingAreasStrategyService` will be switched over to these pure functions and the duplicated `HYSTERESIS`/`requiresHeatingFor`/`requiresHeatingInHysteresisRange` logic deleted in #205 (parent epic #200). The domain layer can't reference the application-layer strategy class, so the duplication is the intended short-lived intermediate state, not a second permanent source of truth.

core/src/main/kotlin/org/agrfesta/sh/api/core/domain/heating/HeatingDecision.kt
AGRFesta: Good catch on the asymmetry, but this is faithful to the current `EconomyAreasSharedHeatingStrategyService`: it already divides the demanding count by the full `areas.size`, and `requiresHeatingFor` already returns `false` for areas with a missing reading or no target — so undecidable areas dilute the ratio today. #204's contract is a byte-for-byte port, so I'm intentionally preserving it here. Whether undecidable areas should be excluded from the denominator is a genuine ECONOMY-semantics change, so I've logged it as a dedicated follow-up (including a characterization test for the current diluting behaviour) rather than slipping it into the port.

core/src/main/kotlin/org/agrfesta/sh/api/core/domain/heating/HeatingDecision.kt
AGRFesta: Agreed this rounding is lossy and the exact cross-multiplication is the better formulation. It's deliberately preserved here only because the current `EconomyAreasSharedHeatingStrategyService` does the same scale-2 `HALF_UP` round before the `>=` comparison, and #204 is a strict byte-for-byte port — switching to the exact comparison changes the decision at boundary thresholds. I've logged it as a dedicated behaviour-change follow-up (with a boundary test) to land alongside/after the #205 wiring rather than silently inside the port.

core/src/main/kotlin/org/agrfesta/sh/api/core/domain/heating/HeatingDecision.kt
AGRFesta: 🤖 **Reviewer**

Verified and conceding. The pure `HYSTERESIS` and `requiresHeating()` here are verbatim equivalents of `AbstractSharedHeatingAreasStrategyService.HYSTERESIS` (`SharedHeatingAreasStrategyService.kt:34`) and `requiresHeatingFor` / `requiresHeatingInHysteresisRange` (`:63-93`), with the only difference being that the impure version reads status via `Either` (failed fetch → `temp <= target`) while the snapshot collapses that into `ActuatorStatus.UNDEFINED`. Agreed this is an acceptable short-lived transitional state given the domain layer can't reference the application-layer strategy class, and the duplication is tracked for removal in #205 (epic #200). Resolving.

core/src/main/kotlin/org/agrfesta/sh/api/core/domain/heating/HeatingDecision.kt
AGRFesta: 🤖 **Reviewer**

Confirmed against the source: `EconomyAreasSharedHeatingStrategyService.internalHandleHeatingFor` divides `areasToHeat.size` by the full `areas.size` (`:55-58`) and `requiresHeatingFor` returns `false` for a missing reading or absent target — so undecidable areas already dilute the ratio today. The port faithfully preserves that, and treating whether to exclude them from the denominator as a separate ECONOMY-semantics change (with a characterization test) is the right call. Resolving as a faithful port; the behaviour question lives in the logged follow-up.

core/src/main/kotlin/org/agrfesta/sh/api/core/domain/heating/HeatingDecision.kt
AGRFesta: 🤖 **Reviewer**

Confirmed: the original `EconomyAreasSharedHeatingStrategyService` does the same scale-2 `HALF_UP` divide before the `>=` comparison (`:56-59`), so the boundary-flip behaviour is pre-existing and the port reproduces it exactly. Agreed the exact cross-multiplication is the better formulation and that switching to it changes decisions at boundary thresholds — correct to land it as a dedicated behaviour-change follow-up with a boundary test rather than inside the port. Resolving.

AGRFesta:
🤖 **Reviewer** — review summary (round 2, sha `64f7b6d3cf8f75f0b377056c2369bde17e92f353`)

Re-review of #204 (delta 92e6c5d..64f7b6d). The only code change since round 1 is the Gemini-suggested switch to the idiomatic toBigDecimal() in decideEconomy — pure, no behaviour change. All three round-1 reviewer threads were rejected with the same well-grounded justification: #204 is a strict port of the existing strategy logic and the questioned behaviours (hysteresis duplication, denominator dilution, lossy scale-2 rounding) are faithful to AbstractSharedHeatingAreasStrategyService / EconomyAreasSharedHeatingStrategyService, with the genuine changes deferred to tracked follow-ups (#205, epic #200). I verified each claim against the source and concede all three. No new findings in the delta.

PR #212, opened 2026-06-20
issue #205: rewire EvaluateHeatingStateService to snapshot + command,…

AGRFesta:
… remove I/O from strategies


core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/EvaluateHeatingStateService.kt
```diff
@@ -52,11 +56,25 @@ class EvaluateHeatingStateService(
...
+                val heaterStatus = heater.getActuatorStatus().getOrElse { ActuatorStatus.UNDEFINED }
+                val snapshots = areaList.map { area ->
+                    HeatableAreaSnapshot(
+                        areaId = area.uuid,
+                        currentTemperature = area.getCurrentTemperature().getOrNull(),
+                        targetTemperature = area.getCurrentTargetTemperature(currentTime),
+                        heaterStatus = heaterStatus
+                    )
+                }
+                when (decide(snapshots)) {
```
AGRFesta: 🤖 **Reviewer** — _severity: low · confidence: medium_

Observability regression. The old strategy services logged every ON/OFF decision (and which areas drove it) and logged temperature-fetch failures. Making the strategies pure is the right call, but no logging was relocated to this orchestrator, so heating decisions are now completely silent in production. Additionally, `area.getCurrentTemperature().getOrNull()` (line 68) discards the `Left` failure that previously produced an `Unable to fetch current temperature for Area` error log. Consider restoring minimal operational visibility here, e.g. capture `val command = decide(snapshots)` and log it with the heater id and snapshots, and log when a sensor reading is unavailable. Keeps the domain pure while preserving debuggability.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/heating/HeatingStrategySelector.kt
```diff
@@ -0,0 +1,62 @@
...
+                    PropertyRepositoryError ->
+                        logger.error("Unable to fetch $HEATING_STRATEGY_KEY property. Falling back to default.")
+                }
+                defaultStrategy
+            },
+            ifRight = { it.value.toSharedHeatingStrategyOrDefault() }
+        )
+
+    private fun String.toSharedHeatingStrategyOrDefault(): SharedHeatingStrategy = try {
+        SharedHeatingStrategy.valueOf(uppercase())
```
AGRFesta: 🤖 **Reviewer** — _severity: low · confidence: low_

`valueOf(uppercase())` uses the default locale and does not trim. The locale concern is currently harmless (the `SharedHeatingStrategy` names COMFORT/ECONOMY contain no locale-sensitive characters, e.g. no `i` for the Turkish-locale case) and this is carried over verbatim from the deleted `DynamicSharedHeatingStrategyService`, so it is not a regression. The more practical robustness gain is `trim()` against accidental leading/trailing whitespace in the stored property value: `SharedHeatingStrategy.valueOf(trim().uppercase(Locale.ROOT))`.

AGRFesta:
🤖 **Reviewer** — review summary (round 1, sha `b4ee5720b1fe4412f385cdc196bb27f42957cb48`)

Clean refactor (#205): the EvaluateHeatingStateService now assembles immutable HeatableAreaSnapshots, resolves a pure HeatingDecider via the thin HeatingStrategySelector, and executes the resulting HeaterCommand. Behaviour is preserved 1:1 vs the deleted strategy services (empty->NONE, hysteresis bands, ECONOMY anti-overheat OFF and demand-ratio threshold), areas are grouped by shared heater with heater status fetched once per group, and the dropped guards (both-strategies-missing, areas-not-sharing-heater) are correctly removed since they are now structurally impossible. Test coverage is thorough. Two minor findings: loss of all heating-decision logging / silent swallowing of temperature-fetch failures in the new orchestrator, and a locale-insensitive/untrimmed strategy-name parse in the selector.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/EvaluateHeatingStateService.kt
AGRFesta: 🤖 **Author** — accepted

Restored operational visibility in the orchestrator: a `warn` when a sensor reading is unavailable (recovering the `Left` previously dropped by `getOrNull()`), and an `info` log of each decision with the heater id and snapshots. The decision logic stays pure in `HeatingDecision`.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/heating/HeatingStrategySelector.kt
AGRFesta: 🤖 **Author** — accepted

Added `valueOf(trim().uppercase(Locale.ROOT))`. `trim()` guards against stray whitespace in the stored property value (covered by a new guard test); `Locale.ROOT` is parse hygiene. Agreed the locale case is harmless for the current enum names — taken as defensive consistency.

Issue #211, opened 2026-06-20
ECONOMY demand ratio: should undecidable areas be excluded from the denominator?


AGRFesta:
# ECONOMY demand ratio: should undecidable areas be excluded from the denominator?

**Parent:** #200
**Related:** #204 (introduced the pure decision), #205 (wiring & cleanup)
**Severity:** Medium
**Type:** Behaviour decision (TDD applies if we change it)

## Context

`decideEconomy` (in `core/domain/heating/HeatingDecision.kt`) computes the demand ratio as:

```kotlin
val demandRatio = areas.count { it.requiresHeating() }.toBigDecimal()
    .divide(areas.size.toBigDecimal(), 2, RoundingMode.HALF_UP)
```

The denominator is the **full** `areas.size`. Areas whose data is undecidable —
`currentTemperature == null` (reading unavailable) or `targetTemperature == null`
(no target configured) — can never satisfy `requiresHeating()`, yet they still
count toward the total. They therefore **dilute** the demand ratio and can suppress
heating when sensor data is partially unavailable.

This is **asymmetric** with `decideComfort`, where such areas are simply ignored by
`any { ... }` (one decidable area demanding heat is enough to turn the heater ON,
regardless of how many areas are undecidable).

## Why this is only a question, not a bug (yet)

This behaviour is a **faithful, byte-for-byte port** of the current
`EconomyAreasSharedHeatingStrategyService`, which already divides the demanding
count by the full `areas.size` and already treats areas with missing reading/target
as non-demanding. #204 deliberately preserved it. This issue is about deciding
whether the **inherited semantics are what we want**, not about a regression.

## Decision to make

- **Option A (status quo):** undecidable areas remain in the denominator and dilute
  the ratio. Rationale: a degraded sensor fleet should bias ECONOMY toward *less*
  heating.
- **Option B:** exclude undecidable areas from the denominator, i.e. compute the
  ratio over **decidable** areas only (`current != null && target != null`),
  matching the "ignore undecidable" stance of `decideComfort`. Define behaviour when
  *all* areas are undecidable (ratio over zero → presumably `OFF`).

## Acceptance criteria

- [ ] A deliberate decision (A or B) is recorded.
- [ ] If Option B: `decideEconomy` excludes undecidable areas from the denominator,
      with a defined result when no area is decidable.
- [ ] Either way, a unit test in `HeatingDecisionTest` pins the chosen semantics
[...]

PR #217, opened 2026-06-28
issue #211: compute ECONOMY demand ratio over decidable areas only

AGRFesta:


core/src/main/kotlin/org/agrfesta/sh/api/core/domain/heating/HeatingDecision.kt
```diff
@@ -31,18 +31,29 @@ fun decideComfort(areas: Collection<HeatableAreaSnapshot>): HeaterCommand = when
...
-            .divide(areas.size.toBigDecimal(), 2, RoundingMode.HALF_UP)
-        if (demandRatio >= threshold.value) HeaterCommand.ON else HeaterCommand.OFF
+fun decideEconomy(areas: Collection<HeatableAreaSnapshot>, threshold: Percentage): HeaterCommand {
+    val decidableCount = areas.count { it.isDecidable() }
+    return when {
+        areas.isEmpty() -> HeaterCommand.NONE
+        areas.any { it.isAboveRange() } -> HeaterCommand.OFF
+        decidableCount == 0 -> HeaterCommand.OFF
+        else -> {
+            val demandRatio = areas.count { it.requiresHeating() }.toBigDecimal()
```
AGRFesta: 🤖 **Reviewer** — _severity: low · confidence: medium_

Robustness: the ratio is correct only because of an *implicit* coupling — the numerator `areas.count { it.requiresHeating() }` runs over ALL areas, while the denominator is `decidableCount`. It stays in `[0,1]` solely because `requiresHeating()` happens to return `false` for undecidable snapshots (its `?: return false` guards). If `requiresHeating()` ever gained a branch that returned `true` without both temperatures present, the numerator could exceed `decidableCount` and yield a ratio > 1. Consider making the invariant explicit by filtering once and computing both terms over the same set, e.g. `val decidable = areas.filter { it.isDecidable() }` then `decidable.count { it.requiresHeating() } / decidable.size`. This also documents intent (numerator ⊆ denominator) and removes the third collection pass. Not a current bug — purely defensive.

AGRFesta:
🤖 **Reviewer** — review summary (round 1, sha `76764c10bc42e93b8d1da12cdfb2f1dce3db7838`)

issue #211 — decideEconomy now divides the demand ratio by the count of decidable areas (both temperatures present) and returns OFF when none is decidable, mirroring decideComfort's ignore-undecidable stance. Logic is correct, division-by-zero is guarded, the new tests include a real regression guard (1 decidable + 2 undecidable flips OFF->ON vs the old areas.size denominator), value objects are used, the function stays pure, and the CHANGELOG entry is present. One low-severity robustness note on an implicit numerator/denominator invariant.

core/src/main/kotlin/org/agrfesta/sh/api/core/domain/heating/HeatingDecision.kt
AGRFesta: 🤖 **Reviewer**

Verified on `eb0055e`. `decidable` is now computed once via `areas.filter { it.isDecidable() }`, and both the numerator (`decidable.count { it.requiresHeating() }`) and denominator (`decidable.size`) run over that same set. The numerator ⊆ denominator invariant is now structural rather than relying on `requiresHeating()`'s implicit `?: return false` guards, so the ratio stays in `[0,1]` even if `requiresHeating()` later gains a branch that returns `true` — and the third collection pass is gone. Resolves the finding.

AGRFesta:
🤖 **Reviewer** — review summary (round 2, sha `eb0055ea656b91d92ddf42d918e186cf412b6eed`)

Re-review of the fixup delta (76764c1..eb0055e). The single open thread — the implicit numerator/denominator coupling — is resolved: decideEconomy now filters once into `decidable` and computes both the numerator (`decidable.count { requiresHeating() }`) and denominator (`decidable.size`) over that same set, making numerator ⊆ denominator structural (ratio guaranteed in [0,1]) and removing the redundant collection pass. Behaviour is unchanged, so the existing regression tests stay valid. No new findings in the delta.

Issue #214, opened 2026-06-20
Follow-up: consistent failure-detail observability in the heating path


AGRFesta:
# Follow-up: consistent failure-detail observability in the heating path

**Type:** Improvement / Observability
**Origin:** Deferred from PR #213 (issue #206 — *Remove I/O from the `Area` domain model*)
**Severity:** Low

## Summary

PR #213 was a behaviour-preserving refactor: it removed I/O from the `Area` domain and moved
snapshot assembly into `EvaluateHeatingStateService`. During review, three observability
*enhancements* were raised that go **beyond** restoring prior behaviour. They were deliberately
deferred so the refactor would not ship a *partial* observability pass — better to do a single,
consistent pass across all failure paths in a dedicated change.

> Note: two related findings were **already addressed** in PR #213 because they were genuine
> regressions, not enhancements:
> - per-sensor read failure severity restored from `warn` back to `error`;
> - "Data inconsistency … not found in persisted records" error logging restored for devices
>   referenced by an area but absent from the device registry.
>
> This follow-up covers only the *new* enhancements below.

## Scope

All in `core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/EvaluateHeatingStateService.kt`.

1. **Include the failure object in fetch-failure logs.** Several `onLeft` / `ifLeft` branches log a
   message but discard the actual `Failure` value (`$it`). Add the failure detail to:
   - the device-fetch failure log (`"Device fetch failed, skipping heating evaluation."`);
   - the area-fetch failure log (`"Area fetch failed, skipping heating evaluation."`);
   - the per-sensor read failure log (`"Unable to fetch readings by sensor '…'."`).

2. **Log heater-status fetch failures.** `heater.getActuatorStatus()` currently degrades to
   `ActuatorStatus.UNDEFINED` silently via `getOrElse`. Replace with a `fold` that logs a warning
   with the failure detail before falling back to `UNDEFINED`.

   ⚠️ There is a guard test asserting the *silent* degradation
   (`execute() maps a heater status fetch failure to UNDEFINED in the snapshot`). It must be
   updated to still assert the `UNDEFINED` outcome while allowing/expecting the new log.

## Acceptance criteria

- [ ] Every failure path in the heating evaluation logs the underlying `Failure` value, not just a
      static message.
- [ ] A heater-status fetch failure is logged (warn) and still degrades to `ActuatorStatus.UNDEFINED`.
[...]

Issue #203, opened 2026-06-19
Fix `SwitchBotMiniHub` reporting the wrong provider


AGRFesta:
# Fix `SwitchBotMiniHub` reporting the wrong provider

**Parent:** #200
**Severity:** Medium (latent data bug, isolated)
**Depends on:** nothing — can ship first, independently
**Type:** Bug fix (TDD applies)

## Summary

`SwitchBotMiniHub` hard-codes its provider to `Provider.NETATMO`, even though it is a SwitchBot
device built by `SwitchBotDevicesFactory`. This is almost certainly a copy/paste bug and exactly the
kind of defect that hides behind the identity/behaviour conflation described in the parent issue.

This sub-issue is a self-contained warm-up: it removes a real bug and establishes a clean first TDD
cycle, with no dependency on the larger refactor.

## Evidence

- `providers/.../switchbot/devices/SwitchBotMiniHub.kt`:
  ```kotlin
  override val provider: Provider = Provider.NETATMO   // should be SWITCHBOT
  ```
- `providers/.../switchbot/devices/SwitchBotDevicesFactory.kt` builds a `SwitchBotMiniHub` for any
  featureless device, so every SwitchBot hub is currently mislabelled as Netatmo.

## Phase 0 — proposed test list

1. `SwitchBotMiniHub` exposes `provider == Provider.SWITCHBOT`.
   *(RED today: returns `NETATMO`.)*
2. Guard test: `SwitchBotDevicesFactory.createDevice(device)` for a featureless device returns a
   driver whose `provider == Provider.SWITCHBOT`.
   *(Locks the bug at the real construction site.)*

Minimal GREEN increment: change the constant in `SwitchBotMiniHub`.

## Acceptance criteria

- [ ] `SwitchBotMiniHub.provider == Provider.SWITCHBOT`.
- [ ] Covered by a unit test and a factory guard test.
- [ ] `CHANGELOG.md` `Fixed` entry under `## [Unreleased]` (observable behaviour change).

## Notes

Follow `docs/TDD_WORKFLOW.md` (bug fix). Smallest, fully-specified slice — good first RED.

PR #218, opened 2026-06-28
issue #214: include failure detail in heating logs and log heater-sta…

AGRFesta:
…tus fetch failures

### DIFF
--- a/core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/EvaluateHeatingStateService.kt
+++ b/core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/EvaluateHeatingStateService.kt
@@ -1,6 +1,5 @@
 package org.agrfesta.sh.api.core.application.usecases
 
-import arrow.core.getOrElse
 import org.agrfesta.sh.api.core.application.ports.inbounds.EvaluateHeatingStateUseCase
 import org.agrfesta.sh.api.core.application.ports.outbounds.TimeProvider
 import org.agrfesta.sh.api.core.application.ports.outbounds.areas.AreasWithDevicesRepository
@@ -54,8 +53,8 @@ class EvaluateHeatingStateService(
 
     override fun execute() {
         if (!isEnabled()) return
-        val deviceRecords = devicesRepository.getAll().onLeft { _ ->
-            logger.error("Device fetch failed, skipping heating evaluation.")
+        val deviceRecords = devicesRepository.getAll().onLeft {
+            logger.error("Device fetch failed, skipping heating evaluation: $it")
         }.getOrNull() ?: return
         val devicesRegistry = deviceRecords.mapNotNull { record ->
             val factory = mappedDevicesFactories[record.provider]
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/EvaluateHeatingStateService.kt
```diff
@@ -130,8 +136,8 @@ class EvaluateHeatingStateService(
     private fun ResolvedHeatableArea.currentTemperature(): Temperature? {
         val readings = sensors.mapNotNull { sensor ->
             sensor.fetchReadings().fold(
-                ifLeft = { _ ->
-                    logger.error("Unable to fetch readings by sensor '${sensor.uuid}'.")
+                ifLeft = {
+                    logger.error("Unable to fetch readings by sensor '${sensor.uuid}': $it")
```
AGRFesta, line 140: 🤖 **Reviewer** — _severity: low · confidence: high_

Log-level inconsistency in the heating path. After this change the code follows a sound implicit convention: failures that **abort the whole evaluation** log at `error` (device fetch L57, area fetch L72), while failures that **degrade one value and let the evaluation continue** log at `warn` (heater-status fetch → UNDEFINED at L81, and 'unable to determine current temperature' at L147). A sensor `fetchReadings()` failure here is the same kind of tolerated degradation — the reading maps to `null`, is filtered out by `mapNotNull`, and `averageTemperature()` carries on — yet it logs at `error`. Given the branch's stated goal is *consistent* failure-detail observability, this is the one remaining outlier: consider `warn` to match L81/L147. (The gemini bot flagged the same on L142.) Note this is a judgment call, not a documented rule — there is no logging-level convention in `docs/`; the call here is internal consistency with the heater-status path you just added. The 'Data inconsistency' device-not-found logs (L109/L121) also degrade-and-continue but represent persistent misconfiguration rather than a transient fetch failure, so keeping those at `error` is defensible — the fetch-failure pair is the inconsistency worth resolving.

AGRFesta:
🤖 **Reviewer** — review summary (round 1, sha `da446b16fb9ab87aa5cd9bb9cb5ca5ca35c2c122`)

Reviewed the #214 observability change (latest commit) plus the stacked refactors it sits on (#203-#211). The refactors are guarded by ArchUnit and HeatingDecisionTest and are settled; focus here is the heating-path logging. One semantic finding: the sensor fetchReadings() failure logs at error while the analogous heater-status fetch failure (and the 'unable to determine current temperature' fallback) log at warn, even though both are tolerated degradations that continue the evaluation. That inconsistency is the one thing this 'consistent failure detail observability' branch arguably still leaves open. Changelog and API docs are correct; tests are behavior-focused with guards for the degradation paths.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/EvaluateHeatingStateService.kt
AGRFesta: 👤 **Author** — decision: rejected

Thanks — legitimate consistency observation, but keeping `error` here as a deliberate policy call.

The branch follows a coherent severity policy along the **I/O-fetch vs value-fallback** axis, not the *abort vs continue* one:
- **`error`** — an I/O fetch failed: device-fetch (L57), area-fetch (L72), sensor `fetchReadings()` (L140). These signal something genuinely wrong (DB error, unreachable hardware).
- **`warn`** — a value degraded to a fallback: heater-status → `UNDEFINED` (L81), no-temperature (L147), multiple heaters.

Under this axis the sensor read sits with the other fetch failures, so it's internally consistent, not drift. An unreachable sensor is a hardware fault we want surfaced at `error`, distinct from "couldn't compute a temperature" (`warn`). This also matches the deliberate restoration in #213 (which #214 frames as the correct level) and the #214 scope, which only asked to add the failure detail (`$it`) to this log, not to change its level.

Re: the gemini reference to "reserve `error` for critical actuation failures" — there is no such documented rule in `docs/`, so I'm not treating it as binding. Leaving as `error`.

Issue #221, opened 2026-06-30
Make `model` the single source of truth


AGRFesta:
# [Device model refactor 2/3] Switch: make `model` the single source of truth

Part of #199. **Switch** step. Depends on **1/3** being deployed and the re-sync gate
(`model IS NULL` count = 0) having passed.

## Goal

Make the persisted device **model** the single discriminator and the single source of truth
for roles. Introduce `DevicePrototype` + a catalog, switch the driver factory and the read
path onto `model`, and collapse the belt-and-suspenders role check. After this, `features`
is still stored but **drives no logic**.

## Why no migration / no DB change here

Pure code change. It only relies on `model` already being populated (guaranteed by the
operational gate of 1/3). Keeping it migration-free is deliberate: it is the middle of the
expand → switch → contract sequence, where no instance may yet depend on `features` being
gone.

### Core — prototype + catalog

- `DevicePrototype` (client-free descriptor: `model`, `provider`, `driverType`).
- `DeviceModelCatalog(prototypes: List<DevicePrototype>)` — `prototypeOf(model)`.
- Roles resolution. **Open decision (carried from #199):**
  - **Recommended here: explicit `roles` on the prototype + a contract test** asserting
    `roles` ⇔ implemented capability interfaces. Reads like domain, debuggable, no runtime
    reflection in a read path.
  - Alternative: derive `roles` from `driverType` via plain Java
    `Class.isAssignableFrom` (no `kotlin-reflect`). Fewer declarations, but roles become an
    invisible projection of the driver type hierarchy.
  - Decide explicitly in this issue before implementing.

### Provider — unconditional prototypes

- `SwitchBotMeterPrototype`, `SwitchBotHubPrototype`, Netatmo prototype, registered as
  **unconditional** `@Component`s (NOT `@ConditionalOnSwitchBot`). The jar is on the
  classpath regardless; only credentials are conditional. This keeps the read path alive
  when the provider is disabled.

### Switch the discriminator and the read path

- `providers/.../switchbot/devices/SwitchBotDevicesFactory.kt` — switch on
  `record.model.value`, **not** on `dto.features.isEmpty()`. This closes the latent bug:
  today `Meter` / `MeterPlus` / `WoIOSensor` all map to `{SENSOR}` and cannot be told apart.
- `core/.../usecases/AssignSensorToAreaService.kt`,
  `AssignActuatorToAreaService.kt` — resolve roles from
[...]

Issue #220, opened 2026-06-30
Introduce 'model'


AGRFesta:
# [Device model refactor 1/3] Expand: introduce `model`, persist it, break nothing

Part of #199. **Expand** step of the expand → switch → contract sequence.

## Goal

Stop discarding the provider's device model and start persisting it as a new, nullable
`model` column — **without changing any read path or behaviour yet**. After this ships, a
re-sync repopulates `model` on every existing row, which is the precondition for the next
sub-issue.

## Why this must be its own deploy

`features → model` is **not invertible** (`Meter` / `MeterPlus` / `WoIOSensor` all collapse
to `{SENSOR}`), and a Flyway migration **cannot call the provider** to recover the model —
it runs at startup, outside the application context. The model can only be repopulated by a
**live application performing a re-sync**. Hence this additive step ships first, on its own.

### Database — migration `V1_1_2__add_model_column_to_device.sql` (additive)

```sql
ALTER TABLE smart_home.device ADD COLUMN model TEXT NULL;

-- Best-effort backfill of the only deterministic case, to shrink the NULL window:
UPDATE smart_home.device SET model = 'switchbot/Hub Mini'
  WHERE provider = 'SWITCHBOT' AND features = '{}';

-- Sensors stay NULL on purpose: features cannot distinguish the model → the re-sync fills them.
```

### Code — dual-write (`features` stays written and read exactly as today)

- `providers/.../switchbot/SwitchBotModel.kt` — expose the model name from
  `SwitchBotDeviceType` (e.g. a `model: String`) alongside `features`.
- `core/.../domain/devices/` — add `@JvmInline value class DeviceModel(val value: String)`;
  add `model: DeviceModel` to `ProviderDeviceData`.
- `providers/.../switchbot/SwitchBotService.kt` — **stop discarding `deviceType`** in
  `getAllDevices()`: today it does `features = it.deviceType.features` and drops the model;
  also map `model = "switchbot/${...}"`.
- `providers/.../netatmo/NetatmoService.kt` — assign a constant `model`
  (e.g. `"netatmo/Smarther"`).
- `core/.../domain/devices/Device.kt` — add nullable `model`.
- `persistence/.../jdbc/entities/DeviceEntity.kt`, `DevicesJdbcRepository.kt`,
  `DevicesJdbcAdapter.kt`, `EntitiesMother.kt` — add `model`, write it in INSERT/UPDATE,
  read it back. **Keep writing and reading `features` unchanged.**

[...]

Issue #199, opened 2026-06-17
Refactor the Device model: separate Prototype, Record and Driver


AGRFesta:
# Refactor the Device model: separate Prototype, Record and Driver

## Motivation

The current model splits a device into two overlapping concepts:

- `Device` — the persisted per-instance state, including a `Set<DeviceFeature>`.
- `DeviceDriver` — the live, operable object, whose capabilities are expressed by the
  interfaces it implements (`Sensor`, `Actuator`/`OnOffActuator`, `BatteryPowered`,
  `Inspectable`).

This creates **two sources of truth for a device's roles**:

1. The persisted `Device.features` set.
2. The capability interfaces the driver implements.

Symptoms in the codebase today:

- `FetchSensorReadingsService` carries a belt-and-suspenders guard
  `!device.features.contains(SENSOR) || driver !is Sensor` — two checks for one fact.
- `docs/domain/DEVICES.md` has to introduce and *manually enforce* a "Role coherence
  invariant" between `Device.features` and the driver interfaces. An invariant you must
  enforce by hand across two representations is the duplication smell itself.
- `SwitchBotDevicesFactory` selects the concrete driver via `dto.features.isEmpty()`
  (empty ⇒ Hub, otherwise Meter). But `Meter`, `MeterPlus` and `WoIOSensor` all map to
  `{SENSOR}`, so `features` **cannot** distinguish two sensor models that need different
  drivers. This is a latent bug: the discriminator is `features`, when it should be the
  device model.

## Proposal

Promote the **device model** to the single discriminator and single source of truth, and
split the model into **three** concepts, each with a distinct reason to exist.

| Concept | What it is | Lives | Needs the provider client? |
|---|---|---|---|
| `DeviceRecord` (was `Device`) | persisted per-instance facts: `uuid`, `provider`, `deviceProviderId`, `name`, `status`, `model` | DB | no |
| `DevicePrototype` (new) | pure, client-free descriptor of a model: which roles/capabilities it has, which driver class it builds | classpath, always | **no** |
| `Device` (was `DeviceDriver`) | the live, operable object; `Sensor`/`Actuator`/`BatteryPowered`/`Inspectable` extend it | runtime | yes |

Key constraint preserved: the **prototype must be resolvable without a live client and
even when the originating provider is disabled**, because role queries happen *before* a
driver exists (lean device list, area-assignment validation, SQL-equivalent filtering).

### Model sketch

Core — persisted record (no more `features`):

```kotlin
[...]

PR #224, opened 2026-07-02
issue #221: make device model the single source of truth for drivers …

AGRFesta:
…and roles


providers/src/main/kotlin/org/agrfesta/sh/api/providers/switchbot/devices/SwitchBotDevicesFactory.kt
```diff
@@ -15,11 +16,10 @@ class SwitchBotDevicesFactory(
 ) : ProviderDevicesFactory {
     override val provider = Provider.SWITCHBOT
 
-    override fun createDevice(dto: Device): DeviceDriver {
-        return if (dto.features.isEmpty()) {
-            SwitchBotMiniHub(dto.uuid, dto.deviceProviderId)
+    override fun createDevice(record: Device): DeviceDriver =
+        if (record.model?.value == SwitchBotDeviceType.HUB_MINI.model) {
```
AGRFesta, line 20: 🤖 **Reviewer** — _severity: medium · confidence: medium_

Driver-identity regression for not-yet-resynced SwitchBot hubs. The discriminator is now `record.model?.value == HUB_MINI.model`: for a legacy hub whose `model` is still NULL (the nullable-model window introduced by #220), this evaluates false and the factory builds a `SwitchBotMeter` (a `Sensor`) instead of a `SwitchBotMiniHub`. Previously the hub was identified by `features.isEmpty()` and built correctly. This is amplified by the guard removal in `FetchSensorReadingsService.collectReadings` (now `if (driver !is Sensor) return`, no feature gate): a misbuilt hub will pass that check and its `fetchReadings()` will hit the SwitchBot status API for a device that has no readings.

I am intentionally not endorsing the features-fallback fix — DEVICES.md now states roles/drivers are resolved 'never from the persisted features', so re-adding a features branch would betray the refactor's stated intent. The design-consistent mitigations are either (a) confirm the operational gate — a successful sync has repopulated `model` for every persisted device before this deploys — or (b) a one-off backfill / a startup log that surfaces any lingering NULL-model rows. Please confirm which guarantees the gate, since the code itself no longer does.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/devices/DeviceModelCatalog.kt
```diff
@@ -0,0 +1,18 @@
...
+
+    private val byModel: Map<DeviceModel, DevicePrototype> = prototypes.associateBy { it.model }
+
+    fun prototypeOf(model: DeviceModel): DevicePrototype? = byModel[model]
+
+    /**
+     * Roles the given [model] can play in an area, resolved from its prototype. Returns an empty set
+     * when [model] is null (device not yet re-synced) or unknown to the catalog.
+     */
+    fun rolesOf(model: DeviceModel?): Set<DeviceFeature> = model?.let { prototypeOf(it)?.roles }.orEmpty()
```
AGRFesta, line 24: 🤖 **Reviewer** — _severity: medium · confidence: medium_

Roles-side of the same NULL-model window. `rolesOf(model)` returns `emptySet()` for a null or unregistered model, and three consumers now depend on it: `AssignSensorToAreaService` / `AssignActuatorToAreaService` will reject a not-yet-resynced device with `NotASensor` / `NotAnActuator` even when it genuinely is one (e.g. a legacy Netatmo Smarther), and `SnapshotSensorHistoryService.filter { SENSOR in catalog.rolesOf(it.model) }` will silently skip null-model sensors from history until the next sync. Same root cause and same design tension as the factory finding — flagging it separately because it is a distinct failure mode (silent exclusion / user-visible 400) on a different code path. If the migration gate is confirmed this is a non-issue; if not, these are the observable symptoms.

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- Distinguish SwitchBot sensor models (Meter/MeterPlus/WoIOSensor) when building device drivers. (#199)
```
AGRFesta: 🤖 **Reviewer** — _severity: low · confidence: high_

Changelog entry is both inaccurately worded and attributed to the wrong issue. (1) 'Distinguish SwitchBot sensor models (Meter/MeterPlus/WoIOSensor) when building device drivers' — all three prototypes declare `driverType = SwitchBotMeter::class`, so no driver-level distinction actually results; the only observable driver change is hub detection switching from `features.isEmpty()` to a model match. (2) The entry is tagged `(#199)` but this PR implements issue #221 (#199 is the epic / base branch). CLAUDE.md's convention is the specific issue number in parentheses. Note also that the sibling sub-issue #220 added no changelog entry at all (treated as internal refactor without a contract change) — worth deciding whether #221 warrants one, and if so, phrase it around the real observable change.

providers/src/test/kotlin/org/agrfesta/sh/api/providers/DevicePrototypesContractTest.kt
```diff
@@ -0,0 +1,86 @@
...
+        homeId = "home",
+        roomId = "room"
+    )
+
+    private val factories: Map<Provider, ProviderDevicesFactory> = listOf(
+        SwitchBotDevicesFactory(mockk<SwitchBotDevicesClient>()),
+        NetatmoDevicesFactory(netatmoConfig, mockk<NetatmoClient>(), mockk<TimeProvider>())
+    ).associateBy { it.provider }
+
+    private val prototypes: List<DevicePrototype> = listOf(
```
AGRFesta, line 50: 🤖 **Reviewer** — _severity: low · confidence: medium_

Completeness gap in the contract test. The two `@TestFactory` blocks iterate the hardcoded `prototypes` list, so they verify each *listed* prototype agrees with its factory/driver — but nothing guarantees the list is exhaustive against the models a provider can actually emit. If a new `SwitchBotDeviceType` (or Netatmo model) is added later without a matching prototype, `rolesOf` returns empty and the factory falls through to `SwitchBotMeter`, and no test fails. Consider a guard that asserts every persistable provider model (e.g. every relevant `SwitchBotDeviceType` entry) has a registered prototype, so the SSOT can't silently develop a hole.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/devices/DeviceModelCatalog.kt
```diff
@@ -0,0 +1,18 @@
+package org.agrfesta.sh.api.core.application.devices
+
+import org.agrfesta.sh.api.core.application.ports.outbounds.devices.DevicePrototype
+import org.agrfesta.sh.api.core.domain.devices.DeviceFeature
+import org.agrfesta.sh.api.core.domain.devices.DeviceModel
+
+class DeviceModelCatalog(prototypes: List<DevicePrototype>) {
+
+    private val byModel: Map<DeviceModel, DevicePrototype> = prototypes.associateBy { it.model }
```
AGRFesta, line 9: 🤖 **Reviewer** — _severity: low · confidence: low_

`prototypes.associateBy { it.model }` silently drops earlier entries on a duplicate `model`, so two prototypes claiming the same model would let the last-registered one win with no signal — a quiet way to corrupt the single source of truth. A fail-fast (e.g. `require` no duplicate models, or `groupBy` + check) would make a misconfiguration surface at startup rather than as a mysterious role mismatch.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/AssignActuatorToAreaService.kt
```diff
@@ -33,10 +36,11 @@ class AssignActuatorToAreaService(
                 devicesRepository.getDeviceById(deviceId)
                     .mapLeft { it.toActuatorFailure() }
                     .flatMap { device ->
-                        if (device.isActuator()) {
+                        val roles = catalog.rolesOf(device.model)
+                        if (DeviceFeature.ACTUATOR in roles) {
                             actuatorsAssignmentsRepository.assign(areaId, deviceId)
                         } else {
-                            NotAnActuator(device.uuid, device.features).left()
+                            NotAnActuator(device.uuid, roles).left()
```
AGRFesta, line 43: 🤖 **Reviewer** — _severity: low · confidence: medium_

Minor semantic drift: `NotAnActuator` / `NotASensor` still carry a field named `features`, but the assign services now populate it from `catalog.rolesOf(device.model)` (catalog roles), not from `device.features`. For an unknown/null model this now reports `emptySet()` rather than the device's real persisted features. It is not client-visible (the controller emits a fixed message and never serializes this field), so this is internal-only — but after this refactor the name is a misnomer; consider renaming to `roles` for clarity, or dropping the field if only `deviceId` is actually consumed. (Same applies to `AssignSensorToAreaService` / `NotASensor`.)

AGRFesta:
🤖 **Reviewer** — review summary (round 1, sha `087e307afbc0acd1fd25eb9cb5e21f90dc2b9242`)

Review of #221 (make the persisted model the single source of truth for drivers and roles). The refactor is well-scoped and well-tested: DeviceModelCatalog + DevicePrototype cleanly replace the features-based branching, the DevicePrototypesContractTest pins the residual model->driver and roles<->capability couplings, and DEVICES.md is updated to match. My findings cluster around one design axis — the behaviour of not-yet-resynced devices whose persisted model is still NULL — plus a few doc/naming nits. I deliberately do NOT recommend the features-fallback the bot review proposes, because DEVICES.md now states roles are resolved 'never from the persisted features'; the real question for the human is whether the migration gate (a sync has repopulated every model before this ships) is actually guaranteed.

CHANGELOG.md
AGRFesta: 🤖 **Author**

Removed the entry. You're right on both counts: the three sensor models all build `SwitchBotMeter`, so there is no driver-level distinction, and under the migration gate there is no observable behaviour change — it is an internal switch, consistent with sibling #220 which added no changelog entry. (The `(#199)` tag was wrong too — it should have been #221.)

core/src/main/kotlin/org/agrfesta/sh/api/core/application/devices/DeviceModelCatalog.kt
AGRFesta: 🤖 **Author**

Good call. Added a fail-fast `require` in the `DeviceModelCatalog` init block that rejects duplicate models at startup, so a misconfigured SSOT surfaces immediately instead of the last-registered prototype silently winning.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/devices/DeviceModelCatalog.kt
AGRFesta: 🤖 **Author**

Confirmed — this is guarded operationally, not by code, by design. Before 2/3 ships a `POST /devices/synchronizations` runs and `SELECT count(*) FROM smart_home.device WHERE model IS NULL` must be 0, so the NULL-model window is closed before deploy. I'm deliberately not re-introducing a `features` fallback: DEVICES.md now states roles are resolved 'never from the persisted features', and 3/3 makes `model` NOT NULL while dropping `features` from logic — a fallback would be throwaway coupling to exactly what we're removing.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/AssignActuatorToAreaService.kt
AGRFesta: 🤖 **Author**

Fair point — the field now carries catalog roles, not `device.features`, so the name is a misnomer. It is internal-only (never serialized; the controller emits a fixed message), so I'm keeping it out of this PR to stay focused and will rename it on the failure types as a follow-up / fold it into 3/3 when the `features` terminology retires.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/switchbot/devices/SwitchBotDevicesFactory.kt
AGRFesta: 🤖 **Author**

Confirmed — guarded operationally by the migration gate, not by code, by design. Before 2/3 deploys a sync repopulates `model` for every device and `model IS NULL` is verified to be 0, so a legacy hub cannot reach the factory with a NULL model. I'm not re-adding the `features.isEmpty()` fallback: DEVICES.md now says drivers/roles are resolved 'never from the persisted features', and 3/3 makes `model` NOT NULL (a hard DB-level gate) — the fallback would be throwaway coupling, removed again one step later.

providers/src/test/kotlin/org/agrfesta/sh/api/providers/DevicePrototypesContractTest.kt
AGRFesta: 🤖 **Author**

Added. A new guard test asserts every `SwitchBotDeviceType` has a registered prototype, so the catalog cannot silently develop a hole when a new device type is introduced. Netatmo has no model enum to iterate, so it stays covered by the per-prototype contract tests.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/switchbot/devices/SwitchBotDevicesFactory.kt
```diff
@@ -15,11 +16,10 @@ class SwitchBotDevicesFactory(
...
+            SwitchBotMiniHub(record.uuid, record.deviceProviderId)
         } else {
-            SwitchBotMeter(dto.uuid, dto.provider, dto.deviceProviderId, client)
+            SwitchBotMeter(record.uuid, record.provider, record.deviceProviderId, client)
         }
```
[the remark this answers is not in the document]
AGRFesta: 🤖 **Author**

Not applying this (nor the sibling `rolesOf(device)` / `filter { … rolesOf(it) }` suggestions). It re-introduces the `features`-based branching this refactor removes — DEVICES.md now states roles/drivers are resolved 'never from the persisted features'. The not-yet-resynced NULL-model case is handled by the operational migration gate (a sync repopulates every `model`, verified `model IS NULL = 0`, before 2/3 deploys), and 3/3 makes `model` NOT NULL, so the fallback would be throwaway coupling. Same conclusion as the reviewer summary.

Issue #222, opened 2026-06-30
Drop `features`


AGRFesta:
# [Device model refactor 3/3] Contract: drop `features`

Part of #199. **Contract** step. Depends on **2/3** being deployed (no running code reads
`features` for logic).

## Goal

Remove `features` as stored data now that `model` + the catalog are the single source of
truth, while **preserving the `GET /devices` API contract**.

## Why a separate, final deploy

Expand → switch → **contract**. Dropping the `features` column while an old instance still
reads it would break that instance mid-rollout. It is only safe once 2/3 is fully deployed
and no running code depends on `features`.

### Database — migration `V1_1_3__make_model_not_null_and_drop_features.sql`

```sql
ALTER TABLE smart_home.device ALTER COLUMN model SET NOT NULL;
ALTER TABLE smart_home.device DROP COLUMN features;
```

Safe because 1/3 + the re-sync guaranteed `model` is fully populated, and 2/3 removed every
logic dependency on `features`.

### Code

- Remove `features` from `core/.../domain/devices/Device.kt` (and `isSensor()` /
  `isActuator()`), `DeviceEntity.kt`, the JDBC repository/adapter, and the test mothers.
- `core/.../domain/failures/Assignments.kt` — `NotASensor` / `NotAnActuator` carry the
  **roles derived from the catalog** instead of `features`.

### API contract — keep the wire stable (decision to confirm in this issue)

- Keep the JSON field named **`features`** on `DeviceResponse` and
  `DeviceAggregateResponse`, with values `SENSOR` / `ACTUATOR`, now **derived** from the
  catalog rather than read from a stored set. No client breaks.
- Internally this is `roles`; only the serialized field name stays `features`.

## Out of scope

- The `DeviceDriver → Device` / `Device → DeviceRecord` rename. That is a separate,
  optional, isolated step (pure structural refactor, ~50 files + wire vocabulary) tracked on
  its own — see #199. Recommendation stands: low value / high blast radius; defer or skip.

## Testing

Schema + contract change → **TDD workflow applies**.

- `GET /devices` and `GET /devices/{uuid}` still serialize `features` with the same values,
  now derived.
- Migration test / persistence slice confirms `model NOT NULL` and `features` gone.

## Acceptance criteria

- [ ] `V1_1_3` makes `model` NOT NULL and drops `features`.
- [ ] `features` removed from domain, persistence, and mothers.
- [ ] `NotASensor` / `NotAnActuator` carry derived roles.
[...]

PR #225, opened 2026-07-03
issue #222: drop the persisted features column and derive device role…

AGRFesta:
…s from the model

### DIFF
--- a/docs/domain/DEVICES.md
+++ b/docs/domain/DEVICES.md
@@ -81,10 +81,9 @@ For each registered provider:
 The operation is **best-effort**: a failure on a single device (e.g. a DB insert error) does not abort
 the synchronization. The remaining devices are processed independently.
 
-The persisted `model` is **nullable**: rows created before the `model` column existed carry `NULL`
-until the next successful sync repopulates it from the provider snapshot (for every device the provider
-reports, new and updated alike). Detached devices retain their last known `model`, since the provider no
-longer reports them.
+The persisted `model` is **always set** (`NOT NULL`): it is populated from the provider snapshot on
+every sync, for every device the provider reports (new and updated alike). Detached devices retain
+their last known `model`, since the provider no longer reports them.
 
 ---
 
@@ -98,8 +97,8 @@ then exercises the driver's *capabilities* (`Sensor`, `BatteryPowered`, `Inspect
 The persisted **`model`** is the single source of truth: a `DeviceModelCatalog` of `DevicePrototype`s maps
 each `model` to its driver type and its **roles** (`SENSOR` / `ACTUATOR`). Consumers resolve the relevant
 capability from the driver itself (`driver is Sensor`) or the role from the catalog
-(`catalog.rolesOf(model)`) — never from the persisted `features`, which is retained on the record and in the
-API responses but no longer drives any logic.
+(`catalog.rolesOf(model)`). The `features` array in the API responses is a presentation-only projection of
+those roles, computed via `catalog.rolesOf(model)`.
 
 | Consumer | Trigger | Uses the device for |
 |----------|---------|---------------------|
--- a/persistence/src/main/resources/db/migration/V1_1_3__make_model_not_null_and_drop_features.sql
+++ b/persistence/src/main/resources/db/migration/V1_1_3__make_model_not_null_and_drop_features.sql
@@ -1,0 +1,6 @@
+-- Backfill any residual NULL model (legacy/detached rows never re-synced) with a non-catalog
+-- sentinel, so SET NOT NULL is correct-by-construction on any environment (restore/replay/fresh).
+-- 'unknown' is absent from the DeviceModelCatalog, so rolesOf() yields empty features.
+UPDATE smart_home.device SET model = 'unknown' WHERE model IS NULL;
+ALTER TABLE smart_home.device ALTER COLUMN model SET NOT NULL;
+ALTER TABLE smart_home.device DROP COLUMN features;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- Derive the `features` array on `GET /devices*` from the device `model` via the catalog; an unknown
+  model yields empty `features`. Wire shape unchanged. (#222)
+
 ## [1.3.2] - 2026-07-02
 
 ## [1.3.1] - 2026-07-01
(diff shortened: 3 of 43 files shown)


persistence/src/main/resources/db/migration/V1_1_3__make_model_not_null_and_drop_features.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE smart_home.device ALTER COLUMN model SET NOT NULL;
```
AGRFesta, line 5: 🤖 **Reviewer** — _severity: medium · confidence: medium_

Robustness, not a live blocker: the author has run a manual sync that repopulated `model` for currently-reported devices, so the acute deploy failure is mitigated for that data. The residual gap is that a sync does **not** cover `DETACHED` devices — the sync leaves their `model` untouched (`DEVICES.md`: "Detached devices retain their last known `model`"), so a device created before `V1_1_2` and detached since before #220 would still hold `model = NULL`. Relying on "a manual sync was run" as an unencoded precondition for a schema migration is fragile: it doesn't hold on a restore/replay or a fresh environment, and provably misses detached-NULL rows. A guard makes the migration correct by construction and is a no-op when the table is already clean:

```sql
UPDATE smart_home.device SET model = 'unknown' WHERE model IS NULL;
ALTER TABLE smart_home.device ALTER COLUMN model SET NOT NULL;
```

A sentinel like `'unknown'` is not in the catalog, so `rolesOf` returns an empty set (empty `features`) — an acceptable, self-correcting degradation. Quick check to confirm the current DB is clean: `SELECT count(*) FROM smart_home.device WHERE model IS NULL;`.

docs/domain/DEVICES.md
```diff
@@ -98,8 +98,8 @@ then exercises the driver's *capabilities* (`Sensor`, `BatteryPowered`, `Inspect
 The persisted **`model`** is the single source of truth: a `DeviceModelCatalog` of `DevicePrototype`s maps
```
AGRFesta, line 97: 🤖 **Reviewer** — _severity: medium · confidence: high_

This PR makes `model` `NOT NULL` in both the schema (`V1_1_3`) and the domain (`Device.model: DeviceModel`), but lines 84–87 just above still state that the persisted `model` is **nullable** and that rows "carry `NULL` until the next successful sync repopulates it". That paragraph is now contradicted by the code and by the very change described here. Per the CLAUDE.md docs convention (docs describe the current state; history lives in the CHANGELOG), please update or drop the 84–87 paragraph so the file is internally consistent.

AGRFesta:
🤖 **Reviewer** — review summary (round 1, sha `bd096a8e787f6b7f4879a0b398f69a99c48374a6`)

Review of #225 (drop the persisted features column; derive roles from model). The core refactor is clean and well-tested: model is the single source of truth, prototypes are unconditional beans so the read path survives a disabled provider, catalog assembly respects hexagonal boundaries, and DevicePrototypesContractTest guards driver/model drift. Two inline findings: (1) the NOT NULL migration has no backfill for pre-existing NULL model rows (robustness — the author confirmed a manual sync mitigates current data, but sync does not cover DETACHED devices); (2) DEVICES.md still documents model as nullable. Minor, no inline anchor (the release/CHANGELOG lives on the base branch, not in this PR's diff): #222 lands with no [Unreleased] CHANGELOG entry — fine if treated as a contract-neutral internal refactor per the CLAUDE.md exemption, but note features is now derived from model, so an unknown model yields empty features, a potentially observable change worth a Changed line.

persistence/src/main/resources/db/migration/V1_1_3__make_model_not_null_and_drop_features.sql
AGRFesta: 🤖 **Author**

Accepted. The manual-sync precondition is fragile and provably misses detached-NULL rows (a device detached before `V1_1_2` is never re-synced, per DEVICES.md), and it doesn't hold on restore/replay/fresh env. Added the idempotent guard `UPDATE … SET model = 'unknown' WHERE model IS NULL` before `SET NOT NULL`; the sentinel is absent from the catalog so `rolesOf` yields empty features — acceptable, self-correcting. Now correct-by-construction and a no-op on a clean table.

docs/domain/DEVICES.md
AGRFesta: 🤖 **Author**

Accepted. Lines 84–87 were contradicted by the code. Rewrote the paragraph to state `model` is always set (`NOT NULL`), populated from the provider snapshot on sync, with detached devices retaining their last known `model` — dropping the nullable/`NULL` framing per the docs-describe-current-state convention.

AGRFesta: 🤖 **Reviewer**

Verified on `cc8fe32`. DEVICES.md §84-86 now reads "The persisted `model` is **always set** (`NOT NULL`): it is populated from the provider snapshot on every sync … Detached devices retain their last known `model`". The contradictory nullable/`NULL`-until-resync paragraph is gone and the file is internally consistent with the `NOT NULL` schema and the non-nullable `Device.model`. Resolved.

persistence/src/main/resources/db/migration/V1_1_3__make_model_not_null_and_drop_features.sql
AGRFesta: 🤖 **Reviewer**

Verified on `cc8fe32`. The migration now runs `UPDATE smart_home.device SET model = 'unknown' WHERE model IS NULL;` before `ALTER COLUMN model SET NOT NULL;`, so it is correct-by-construction on restore/replay/fresh environments and a no-op on an already-clean table — the fragile manual-sync precondition (which provably missed detached-NULL rows) is no longer relied upon. I also confirmed `'unknown'` is not a catalog model key (real keys are `switchbot/*` and the Netatmo SMARTHER_MODEL), so any backfilled row resolves to empty `features` — the acceptable, self-correcting degradation we discussed. Resolved.

AGRFesta:
🤖 **Reviewer** — review summary (round 2, sha `cc8fe326c6741db8bf9492232bac637471e1776e`)

Re-review of #225 (round 2, delta bd096a8..cc8fe32). All round-1 findings addressed in a single fixup commit, verified clean: (1) DEVICES.md §84-87 now states `model` is always set (`NOT NULL`), consistent with the schema and domain — the nullable/`NULL` framing is gone; (2) V1_1_3 gained the idempotent backfill `UPDATE … SET model = 'unknown' WHERE model IS NULL` before `SET NOT NULL`, making the migration correct-by-construction on restore/replay/fresh envs and a no-op on a clean table — I confirmed `'unknown'` is absent from the real catalog (models are `switchbot/*` and the Netatmo SMARTHER_MODEL), so `rolesOf('unknown')` yields empty features, an acceptable self-correcting degradation; (3) the minor round-1 note is also resolved — a `Changed` CHANGELOG entry now records that `features` is derived from `model` (unknown model → empty features, wire shape unchanged). No new issues in the delta. Both reviewer threads resolved; recommend merge.

PR #227, opened 2026-07-03
issue #201: return a typed failure or a per-heater report from the he…

### LINKED ISSUE #201: `EvaluateHeatingStateUseCase.execute()` returns `Unit`: the safety-critical operation has no error contract
# `EvaluateHeatingStateUseCase.execute()` returns `Unit`: the safety-critical operation has no error contract

**Severity:** High
**Area:** `:core` inbound port + use case, `:app` scheduler
**Type:** Architecture / error handling / operability

## Summary

`EvaluateHeatingStateUseCase` is the only use case whose `execute()` returns `Unit` instead of
`Either<Failure, T>`. It is also the most safety-critical operation in the system: it **drives the
physical heating actuators**. Today every failure (device fetch, area fetch, actuator status, actuator
command) is swallowed with a `logger.error(...)` and an early `return`. The caller can never know
whether the evaluation ran, partially ran, or did nothing.

This directly violates the rules in `docs/ARCHITECTURE.md`:
> Inbound ports "Must return functional types (`Either`)."
> DON'T "use `Either.getOrElse` or force-unwrap results in the service layer to avoid handling errors."

The fact that the only caller is a `@Scheduled` task explains the choice but does not justify it: a
port contract must not depend on who happens to call it.

## Evidence

- `core/application/ports/inbounds/EvaluateHeatingStateUseCase.kt`:
  ```kotlin
  fun execute()   // returns Unit
  ```
- `core/application/usecases/EvaluateHeatingStateService.kt`:
  - `if (!isEnabled()) return`
  - `devicesRepository.getAll().onLeft { logger.error(...) }.getOrNull() ?: return`
  - `areasWithDevicesRepository.getAllAreasWithDevices().onLeft { logger.error(...) }.getOrNull()?... ?: return`
  - all failure information is reduced to log lines; nothing is returned.
- `app/.../schedulers/HeatingControlScheduler.kt` calls `evaluateHeatingState.execute()` and
  discards any notion of success/failure.

## Why this matters

- **No observability of a physical-control operation.** A heater that fails to turn on/off because of
  a provider error is indistinguishable from a successful no-op at the call site.
- **No path to alerting or retry.** Because the outcome is `Unit`, the scheduler cannot decide to
  retry, raise an alert, or surface a degraded state.
- **Inconsistency.** Every other use case (~24 of them) follows the `Either<SealedFailure, T>`
  contract; this one is the lone exception, which erodes the convention the codebase otherwise
  enforces well.

## Proposed direction

1. Introduce a sealed failure contract, e.g. `EvaluateHeatingStateFailure`
   (`HeatingDisabled`/`SkippedNoData` may be modelled as a successful result rather than a failure).
2. Change the port to:
   ```kotlin
   fun execute(): Either<EvaluateHeatingStateFailure, HeatingActionReport>
   ```
   where `HeatingActionReport` captures what happened (heaters evaluated, commands issued, per-heater
   outcome). This pairs naturally with the pure command-based heating decision proposed in #200.
3. The scheduler logs/handles the `Left` (and optionally raises an `Alert`, given the alert domain now
   exists).

## Acceptance criteria

- [ ] `EvaluateHeatingStateUseCase.execute()` returns an `Either` with a use-case-specific sealed
      failure (no `Failure` generic, no `Unit`).
- [ ] Failures from device/area fetch and actuator operations are represented in the result, not only
      logged.
- [ ] `HeatingControlScheduler` consumes the result explicitly.
- [ ] Tests cover: disabled flag, device-fetch failure, area-fetch failure, and a successful
      evaluation producing a report.

## Notes

Follow `docs/TDD_WORKFLOW.md` (behaviour change). Best sequenced **after or alongside** #200, so
the report can express per-heater commands cleanly.

AGRFesta:
…ating evaluation

### DIFF
--- a/app/src/main/kotlin/org/agrfesta/sh/api/schedulers/HeatingControlScheduler.kt
+++ b/app/src/main/kotlin/org/agrfesta/sh/api/schedulers/HeatingControlScheduler.kt
@@ -1,6 +1,9 @@
 package org.agrfesta.sh.api.schedulers
 
 import org.agrfesta.sh.api.core.application.ports.inbounds.EvaluateHeatingStateUseCase
+import org.agrfesta.sh.api.core.domain.heating.ActuationOutcome
+import org.agrfesta.sh.api.core.domain.heating.HeatingEvaluationReport
+import org.agrfesta.sh.api.utils.LoggerDelegate
 import org.springframework.scheduling.annotation.Async
 import org.springframework.scheduling.annotation.Scheduled
 import org.springframework.stereotype.Component
@@ -9,9 +12,26 @@ import org.springframework.stereotype.Component
 class HeatingControlScheduler(
     private val evaluateHeatingState: EvaluateHeatingStateUseCase
 ) {
+    private val logger by LoggerDelegate()
+
     @Scheduled(cron = "0 */15 * * * *")
     @Async
     fun scheduledTask() {
-        evaluateHeatingState.execute()
+        evaluateHeatingState.execute().fold(
+            ifLeft = { logger.error("Heating evaluation failed: $it") },
+            ifRight = ::logFailedActuations
+        )
+    }
+
+    private fun logFailedActuations(report: HeatingEvaluationReport) {
+        when (report) {
+            is HeatingEvaluationReport.Evaluated -> report.heaters.forEach { heater ->
+                val outcome = heater.outcome
+                if (outcome is ActuationOutcome.Failed) {
+                    logger.error("Heater '${heater.heaterId}' failed to actuate ${heater.command}: ${outcome.cause}")
+                }
+            }
+            HeatingEvaluationReport.Skipped -> Unit
+        }
     }
 }
--- a/app/src/test/kotlin/org/agrfesta/sh/api/schedulers/HeatingControlSchedulerTest.kt
+++ b/app/src/test/kotlin/org/agrfesta/sh/api/schedulers/HeatingControlSchedulerTest.kt
@@ -1,0 +1,113 @@
+package org.agrfesta.sh.api.schedulers
+
+import arrow.core.left
+import arrow.core.right
+import ch.qos.logback.classic.Level
+import ch.qos.logback.classic.Logger
+import ch.qos.logback.classic.spi.ILoggingEvent
+import ch.qos.logback.core.read.ListAppender
+import io.kotest.assertions.withClue
+import io.kotest.matchers.shouldBe
+import io.mockk.every
+import io.mockk.mockk
+import org.agrfesta.sh.api.core.application.ports.inbounds.EvaluateHeatingStateUseCase
+import org.agrfesta.sh.api.core.domain.failures.ActuatorOperationFailure
+import org.agrfesta.sh.api.core.domain.failures.HeatingFlagUnavailable
+import org.agrfesta.sh.api.core.domain.heating.ActuationOutcome
+import org.agrfesta.sh.api.core.domain.heating.HeaterActionOutcome
+import org.agrfesta.sh.api.core.domain.heating.HeaterCommand
+import org.agrfesta.sh.api.core.domain.heating.HeatingEvaluationReport
+import org.junit.jupiter.api.AfterEach
+import org.junit.jupiter.api.Test
+import org.slf4j.LoggerFactory
+import java.util.UUID
+
+class HeatingControlSchedulerTest {
+
+    private val evaluateHeatingState: EvaluateHeatingStateUseCase = mockk()
+    private val sut = HeatingControlScheduler(evaluateHeatingState)
+
+    private val logger = LoggerFactory.getLogger(HeatingControlScheduler::class.java) as Logger
+    private val appender = ListAppender<ILoggingEvent>().apply { start() }
+
+    init {
+        logger.addAppender(appender)
+    }
+
+    @AfterEach
+    fun detachAppender() {
+        logger.detachAppender(appender)
+    }
+
+    @Test
+    fun `scheduledTask() logs an error for each Failed outcome in the report`() {
+        // Given — two heaters: one actuated fine, one failed
+        val failure = object : ActuatorOperationFailure {}
+        val failedHeaterId = UUID.randomUUID()
+        every { evaluateHeatingState.execute() } returns HeatingEvaluationReport.Evaluated(
+            listOf(
+                HeaterActionOutcome(UUID.randomUUID(), HeaterCommand.ON, ActuationOutcome.Issued),
+                HeaterActionOutcome(failedHeaterId, HeaterCommand.OFF, ActuationOutcome.Failed(failure))
+            )
+        ).right()
+
+        // When
+        sut.scheduledTask()
+
+        // Then
+        withClue("each failed actuation logs an error naming heater and cause; successes are quiet") {
+            appender.list.count {
+                it.level == Level.ERROR &&
+                    it.formattedMessage.contains(failedHeaterId.toString()) &&
+                    it.formattedMessage.contains(failure.toString())
+            } shouldBe 1
+            appender.list.count { it.level == Level.ERROR } shouldBe 1
+        }
+    }
+
+    @Test
+    fun `scheduledTask() is quiet when the evaluation succeeds without failed actuations`() {
+        // Given
+        every { evaluateHeatingState.execute() } returns HeatingEvaluationReport.Evaluated(
+            listOf(HeaterActionOutcome(UUID.randomUUID(), HeaterCommand.ON, ActuationOutcome.Issued))
+        ).right()
+
+        // When
+        sut.scheduledTask()
+
+        // Then
+        withClue("a successful evaluation with no failed actuation must not log errors") {
+            appender.list.count { it.level == Level.ERROR } shouldBe 0
+        }
+    }
+
+    @Test
+    fun `scheduledTask() is quiet when the evaluation is skipped`() {
+        // Given
+        every { evaluateHeatingState.execute() } returns HeatingEvaluationReport.Skipped.right()
+
+        // When
+        sut.scheduledTask()
+
+        // Then
+        withClue("a skipped evaluation (heating disabled) must not log errors") {
+            appender.list.count { it.level == Level.ERROR } shouldBe 0
+        }
+    }
+
+    @Test
+    fun `scheduledTask() logs the failure when evaluation returns Left`() {
+        // Given
+        every { evaluateHeatingState.execute() } returns HeatingFlagUnavailable.left()
+
+        // When
+        sut.scheduledTask()
+
+        // Then
+        withClue("a failed heating evaluation must be logged as an error by the scheduler") {
+            appender.list.count {
+                it.level == Level.ERROR && it.formattedMessage.contains("HeatingFlagUnavailable")
+            } shouldBe 1
+        }
+    }
+}
--- a/core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/EvaluateHeatingStateService.kt
+++ b/core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/EvaluateHeatingStateService.kt
@@ -1,5 +1,7 @@
 package org.agrfesta.sh.api.core.application.usecases
 
+import arrow.core.Either
+import arrow.core.raise.either
 import org.agrfesta.sh.api.core.application.ports.inbounds.EvaluateHeatingStateUseCase
 import org.agrfesta.sh.api.core.application.ports.outbounds.TimeProvider
 import org.agrfesta.sh.api.core.application.ports.outbounds.areas.AreasWithDevicesRepository
@@ -15,9 +17,16 @@ import org.agrfesta.sh.api.core.domain.areas.AreaDtoWithDevices
 import org.agrfesta.sh.api.core.domain.commons.Temperature
 import org.agrfesta.sh.api.core.domain.devices.ActuatorStatus
 import org.agrfesta.sh.api.core.domain.devices.averageTemperature
-import org.agrfesta.sh.api.core.domain.failures.PropertyRepositoryError
+import org.agrfesta.sh.api.core.domain.failures.ActuatorOperationFailure
+import org.agrfesta.sh.api.core.domain.failures.AreasUnavailable
+import org.agrfesta.sh.api.core.domain.failures.DevicesUnavailable
+import org.agrfesta.sh.api.core.domain.failures.EvaluateHeatingStateFailure
+import org.agrfesta.sh.api.core.domain.failures.HeatingFlagUnavailable
+import org.agrfesta.sh.api.core.domain.heating.ActuationOutcome
 import org.agrfesta.sh.api.core.domain.heating.HeatableAreaSnapshot
+import org.agrfesta.sh.api.core.domain.heating.HeaterActionOutcome
 import org.agrfesta.sh.api.core.domain.heating.HeaterCommand
+import org.agrfesta.sh.api.core.domain.heating.HeatingEvaluationReport
 import org.agrfesta.sh.api.utils.LoggerDelegate
 import org.springframework.stereotype.Service
 import java.time.LocalTime
@@ -51,11 +60,12 @@ class EvaluateHeatingStateService(
         val heater: Heater
     )
 
-    override fun execute() {
-        if (!isEnabled()) return
-        val deviceRecords = devicesRepository.getAll().onLeft {
-            logger.error("Device fetch failed, skipping heating evaluation: $it")
-        }.getOrNull() ?: return
+    override fun execute(): Either<EvaluateHeatingStateFailure, HeatingEvaluationReport> = either {
+        if (isEnabled().bind()) evaluateHeatableAreas().bind() else HeatingEvaluationReport.Skipped
+    }
+
+    private fun evaluateHeatableAreas(): Either<EvaluateHeatingStateFailure, HeatingEvaluationReport> = either {
+        val deviceRecords = devicesRepository.getAll().mapLeft { DevicesUnavailable }.bind()
         val devicesRegistry = deviceRecords.mapNotNull { record ->
             val factory = mappedDevicesFactories[record.provider]
             if (factory == null) {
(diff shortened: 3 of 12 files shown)


core/src/test/kotlin/org/agrfesta/sh/api/core/application/usecases/EvaluateHeatingStateServiceTest.kt
```diff
@@ -83,6 +90,18 @@ class EvaluateHeatingStateServiceTest {
         every { strategySelector.select() } returns decide
     }
 
+    @Test
+    fun `execute() returns Skipped when heating is disabled`() {
```
AGRFesta: 🤖 **Reviewer** — _severity: low · confidence: medium_

Six new `returns X` tests were added as siblings of the pre-existing `does nothing when ...` tests with byte-identical Given blocks (disabled flag, flag fetch failure, missing flag, device fetch failure, area fetch failure, no heatable areas). Each pair exercises the same scenario and differs only in the Then: one asserts the returned `Either`, the other asserts `verify(exactly = 0) { decide(any()) }`.

Per the TDD guidelines this is a Phase-3 refactor left undone: the return value and the absence of actuation are two assertions of one behaviour, and the duplicated Givens are noise that will drift apart on the next change. Suggest merging each pair — keep the `verify(exactly = 0)` side-effect assertion inside the return-asserting test (the "no actuation on failure" safety property must stay pinned). Merging also naturally fixes the now-slightly-misleading old names (`does nothing when HEATING_ENABLED_KEY fetch fails` — it no longer "does nothing", it returns `HeatingFlagUnavailable`). Note the asymmetry too: `does nothing when HEATING_ENABLED_KEY is not a boolean string` (line 176) has no `returns Skipped` counterpart.

core/src/main/kotlin/org/agrfesta/sh/api/core/domain/heating/HeatingEvaluationReport.kt
```diff
@@ -0,0 +1,25 @@
...
+
+/** Per-heater outcome of a heating evaluation run. */
+data class HeaterActionOutcome(
+    val heaterId: UUID,
+    val command: HeaterCommand,
+    val outcome: ActuationOutcome
+)
+
+/** How the command decided for a heater turned out. */
+enum class ActuationOutcome { ISSUED, FAILED, NOT_NEEDED }
```
AGRFesta: 🤖 **Reviewer** — _severity: low · confidence: medium_

`ActuationOutcome.FAILED` records *that* an actuation failed but drops the `ActuatorOperationFailure` cause — it survives only as a `logger.error` inside the core service. Issue #201's acceptance criterion is that actuator failures be "represented in the result, not only logged"; this satisfies the letter, but any future consumer (the alerting path the issue mentions) will need the cause and won't have it.

Consider modelling the outcome as a sealed type (`Issued`, `NotNeeded`, `Failed(cause: ActuatorOperationFailure)`) instead of an enum. Not blocking — the enum is a reasonable first increment if the sealed upgrade is deliberate deferred scope, but it's a data-model decision that is cheaper to make now than after consumers exist.

app/src/main/kotlin/org/agrfesta/sh/api/schedulers/HeatingControlScheduler.kt
```diff
@@ -9,9 +10,12 @@ import org.springframework.stereotype.Component
 class HeatingControlScheduler(
     private val evaluateHeatingState: EvaluateHeatingStateUseCase
 ) {
+    private val logger by LoggerDelegate()
+
     @Scheduled(cron = "0 */15 * * * *")
     @Async
     fun scheduledTask() {
         evaluateHeatingState.execute()
+            .onLeft { logger.error("Heating evaluation failed: $it") }
```
AGRFesta: 🤖 **Reviewer** — _severity: medium · confidence: medium_

The scheduler consumes only the `Left`: a `Right(Evaluated)` report containing `FAILED` actuations completes silently at the call site. The only reason this isn't an observability regression is that `EvaluateHeatingStateService` still logs every actuation failure internally — which means the responsibility the report was introduced to move to the caller hasn't actually moved; the core both reports *and* logs, and the report's content is discarded by its only consumer.

Suggest having the scheduler inspect the report and log (warn/error) when any outcome is `FAILED` — that is where the per-heater report becomes actionable (and the natural seam for the future Alert). The duplicate `logger.error` in `toActuationOutcome` inside the core service can then be dropped or downgraded. If you prefer to keep this increment minimal and defer to the alerting issue, that's a defensible call — say so and I'll resolve.

app/src/test/kotlin/org/agrfesta/sh/api/schedulers/HeatingControlSchedulerTest.kt
```diff
@@ -0,0 +1,50 @@
...
+    init {
+        logger.addAppender(appender)
+    }
+
+    @AfterEach
+    fun detachAppender() {
+        logger.detachAppender(appender)
+    }
+
+    @Test
```
AGRFesta, line 98: 🤖 **Reviewer** — _severity: low · confidence: high_

Only the `Left` path is tested. A companion test for the happy path — `execute()` returns `Right(report)`, assert no ERROR event is appended — would pin the contract that a successful evaluation is quiet, and guard against a future refactor logging the report unconditionally at error level. Cheap to add since the appender scaffolding is already in place.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/EvaluateHeatingStateService.kt
```diff
@@ -96,14 +104,22 @@ class EvaluateHeatingStateService(
...
             }
+        HeatingEvaluationReport.Evaluated(outcomes)
+    }
+
+    private fun Either<ActuatorOperationFailure, Unit>.toActuationOutcome(
+        heater: Heater,
+        command: HeaterCommand
+    ): ActuationOutcome {
+        onLeft { logger.error("Failed to turn heater '${heater.uuid}' $command: $it") }
+        return if (isLeft()) ActuationOutcome.FAILED else ActuationOutcome.ISSUED
```
AGRFesta: 🤖 **Reviewer** — _severity: low · confidence: high_

Style: `toActuationOutcome` inspects the `Either` twice (`onLeft { ... }` then `isLeft()`). A single `fold` is the idiom used everywhere else in this same file (`getActuatorStatus`, `fetchReadings`):

```kotlin
private fun Either<ActuatorOperationFailure, Unit>.toActuationOutcome(
    heater: Heater,
    command: HeaterCommand
): ActuationOutcome = fold(
    ifLeft = {
        logger.error("Failed to turn heater '${heater.uuid}' $command: $it")
        ActuationOutcome.FAILED
    },
    ifRight = { ActuationOutcome.ISSUED }
)
```

AGRFesta:
🤖 **Reviewer** — review summary (round 1, sha `6039ac84b4595f0aa4f9aa84f5a78e5e777f4951`)

Review of #227 (issue #201). The core change is sound: the port now returns `Either<EvaluateHeatingStateFailure, HeatingEvaluationReport>`, the sealed failure contract matches the established pattern in `core/domain/failures/`, the KDoc was updated to describe the new contract, and the changelog entry is present and correctly scoped. All four acceptance-criteria scenarios (disabled flag, device-fetch failure, area-fetch failure, successful evaluation report) are covered by tests. Findings are about test duplication left behind by the migration, the report dropping the actuation failure cause, the scheduler ignoring the successful report (including partial FAILED outcomes), a missing happy-path scheduler test, and a minor style point.

core/src/main/kotlin/org/agrfesta/sh/api/core/domain/heating/HeatingEvaluationReport.kt
AGRFesta: 🤖 **Author** — Accepted. `ActuationOutcome` is now a sealed type (`Issued`, `NotNeeded`, `Failed(cause: ActuatorOperationFailure)`). To keep the domain→ports dependency direction clean, `ActuatorOperationFailure` moved from `application/ports/outbounds/devices` to `core/domain/failures` (still non-sealed on purpose). Fixed in 6a16f60.

app/src/main/kotlin/org/agrfesta/sh/api/schedulers/HeatingControlScheduler.kt
AGRFesta: 🤖 **Author** — Accepted. The scheduler now folds the result: `Left` → error log, `Right` → one error log per `Failed` outcome (heater, command, cause). The duplicate `logger.error` in the core's `toActuationOutcome` was removed — the core only reports, the scheduler owns the logging. Fixed in 6a16f60.

app/src/test/kotlin/org/agrfesta/sh/api/schedulers/HeatingControlSchedulerTest.kt
AGRFesta: 🤖 **Author** — Accepted. Added `scheduledTask() is quiet when the evaluation succeeds without failed actuations` as a green-on-arrival guard. Fixed in 6a16f60.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/EvaluateHeatingStateService.kt
AGRFesta: 🤖 **Author** — Accepted, with one necessary companion change the suggestion misses: `fold` is an exhaustive inlined `when`, and the legacy tests exercise unstubbed `on()`/`off()` on a `mockk(relaxed = true)` heater whose returned `Either` is neither `Left` nor `Right` — `fold` would throw `NoWhenBranchMatchedException`. `toSharedHeaterMockk` now stubs `on()`/`off()` to `Unit.right()` by default. With the sealed outcome the function reduced to a pure `fold(ifLeft = { Failed(it) }, ifRight = { Issued })`. Fixed in 6a16f60.

core/src/test/kotlin/org/agrfesta/sh/api/core/application/usecases/EvaluateHeatingStateServiceTest.kt
AGRFesta: 🤖 **Author** — Accepted. Each pair merged into a single test keeping the `verify(exactly = 0)` safety assertion alongside the result assertion; names updated to `returns X and does not act when ...`; the non-boolean flag case now also asserts `Right(Skipped)`. Fixed in 6a16f60.

app/src/main/kotlin/org/agrfesta/sh/api/schedulers/HeatingControlScheduler.kt
```diff
@@ -9,9 +12,24 @@ import org.springframework.stereotype.Component
...
     fun scheduledTask() {
-        evaluateHeatingState.execute()
+        evaluateHeatingState.execute().fold(
+            ifLeft = { logger.error("Heating evaluation failed: $it") },
+            ifRight = ::logFailedActuations
+        )
+    }
+
+    private fun logFailedActuations(report: HeatingEvaluationReport) {
+        val outcomes = (report as? HeatingEvaluationReport.Evaluated)?.heaters.orEmpty()
```
AGRFesta: 🤖 **Reviewer** — _severity: low · confidence: medium_

Style, non-blocking: `(report as? HeatingEvaluationReport.Evaluated)?.heaters.orEmpty()` opts out of the sealed hierarchy's exhaustiveness. A `when (report) { is Evaluated -> ...; Skipped -> Unit }` would force this call site to be revisited if a new report subtype is ever added, whereas the `as?` cast will silently ignore it. Fine to leave as-is for this increment; flagging it only because the sealed type was introduced precisely to make consumers total.

app/src/main/kotlin/org/agrfesta/sh/api/schedulers/HeatingControlScheduler.kt
AGRFesta: 🤖 **Reviewer**

Verified in 6a16f60: the scheduler folds the result — `Left` logs the evaluation failure, `Right` logs one error per `Failed` outcome naming heater, command and cause. The duplicate `logger.error` in the core's `toActuationOutcome` is gone, so the core now only reports and the caller owns observability, which is exactly the separation the report was meant to enable. Covered by the new `logs an error for each Failed outcome` test.

app/src/test/kotlin/org/agrfesta/sh/api/schedulers/HeatingControlSchedulerTest.kt
AGRFesta: 🤖 **Reviewer**

Verified in 6a16f60: the happy-path guard test asserts zero ERROR events on a successful report, and the additional per-Failed-outcome test pins the count at exactly one error for the failed heater (successes quiet). Both use the existing appender scaffolding.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/EvaluateHeatingStateService.kt
AGRFesta: 🤖 **Reviewer**

Verified in 6a16f60: `toActuationOutcome` is now a pure `fold(ifLeft = { Failed(it) }, ifRight = { Issued })`. The companion change is correct and necessary — `fold` is an exhaustive inlined `when`, so `toSharedHeaterMockk` stubbing `on()`/`off()` to `Unit.right()` (with an explanatory comment) is the right fix for the relaxed-mock Either problem; failure-path tests still override with `.left()` explicitly.

core/src/main/kotlin/org/agrfesta/sh/api/core/domain/heating/HeatingEvaluationReport.kt
AGRFesta: 🤖 **Reviewer**

Verified in 6a16f60: `ActuationOutcome` is a sealed type with `Failed(cause: ActuatorOperationFailure)`, and both failure-path tests now assert the cause is carried in the report. Relocating `ActuatorOperationFailure` to `core/domain/failures` to keep the domain→ports dependency direction clean is the right call, and it stays non-sealed with a KDoc stating why — consistent with the documented outbound-port failure contract pattern. No stale imports of the old location remain.

core/src/test/kotlin/org/agrfesta/sh/api/core/application/usecases/EvaluateHeatingStateServiceTest.kt
AGRFesta: 🤖 **Reviewer**

Verified in 6a16f60: each duplicated pair is merged into a single test keeping the `verify(exactly = 0)` safety assertions alongside the result assertion, names now read `returns X and does not act when ...`, and the previously asymmetric non-boolean flag case pins `Right(Skipped)`. The no-actuation-on-failure safety property remains covered in every merged test.

AGRFesta:
🤖 **Reviewer** — review summary (round 2, sha `6a16f60080be76478760d0953909147c61805704`)

Round 2 re-review of the delta 6039ac8..6a16f60. All five round-1 threads were accepted and fixed in 6a16f60; every fix was verified against the delta and all five threads are resolved. The scheduler now folds the result and owns the failure logging (the core's duplicate logger.error is gone), the happy-path scheduler test plus a per-Failed-outcome logging test were added, toActuationOutcome reduced to a pure fold with the necessary test-fixture stubbing of on()/off(), ActuationOutcome is now a sealed type carrying the ActuatorOperationFailure cause (with the failure interface correctly relocated to core/domain/failures, keeping it deliberately non-sealed per the documented outbound-port failure contract), and the duplicated test pairs were merged with the non-boolean flag case now pinning Right(Skipped). One new low-severity style note on the new scheduler code; nothing blocking. The unmarked gemini-code-assist inline comment (fold suggestion) is addressed by the same fix as thread 3520913681.

app/src/main/kotlin/org/agrfesta/sh/api/schedulers/HeatingControlScheduler.kt
AGRFesta: 🤖 **Author** — Accepted. `logFailedActuations` now pattern-matches exhaustively on the sealed report (`Evaluated` → per-`Failed` error logs, `Skipped -> Unit`), so any future report subtype will fail compilation at this call site instead of being silently ignored. Also added a `is quiet when the evaluation is skipped` guard test covering the `Skipped` branch (plus a detekt line-length fix in the Failed-outcome test's clue). Fixed in 8dc85e0.

AGRFesta: 🤖 **Reviewer**

Verified in 8dc85e0: `logFailedActuations` now uses an exhaustive `when` over the sealed report (`Evaluated` -> per-`Failed` error logs, `Skipped -> Unit`), so a future report subtype breaks compilation here instead of being silently ignored. The new `is quiet when the evaluation is skipped` guard test pins the `Skipped` branch. Exactly the fix suggested.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/EvaluateHeatingStateService.kt
```diff
@@ -96,14 +104,22 @@ class EvaluateHeatingStateService(
...
     }
```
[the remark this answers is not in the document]
AGRFesta: 🤖 **Reviewer**

Addressed in 6a16f60 together with reviewer thread on the same lines: `toActuationOutcome` is now a pure `fold(ifLeft = { Failed(it) }, ifRight = { Issued })` — the `onLeft` + `isLeft()` double inspection is gone (the logging moved to the scheduler, which owns observability). Closing as resolved.

AGRFesta:
🤖 **Reviewer** — review summary (round 3, sha `8dc85e0dd0cb5f7f89f3287158845e9c6e80a32b`)

Round 3 re-review of the delta 6a16f60..8dc85e0. The delta contains exactly the fix for the single open round-2 thread: `logFailedActuations` now pattern-matches exhaustively on the sealed `HeatingEvaluationReport` (`Evaluated` -> per-`Failed` error logs, `Skipped -> Unit`), so any future report subtype fails compilation at this call site instead of being silently dropped by the previous `as?` cast. A `scheduledTask() is quiet when the evaluation is skipped` guard test pins the `Skipped` branch, and the Failed-outcome test's clue was shortened for detekt line length. Verified by running `HeatingControlSchedulerTest` — green. No new findings; all threads are now resolved, including the unmarked gemini-code-assist fold suggestion, which was addressed by the round-1 fix to `toActuationOutcome`. Nothing blocking from the reviewer side.

Issue #195, opened 2026-06-14
Extensions: real channels, new rules, manual ack/snooze


AGRFesta:
# Extensions: real channels, new rules, manual ack/snooze

## Context
Issues #191–#194 deliver the full vertical: battery in cache → automatic alerts → notifications with reminder cadence,
delivered through a persisted channel. This issue is an **umbrella for follow-ups** that extend the same engine
without changing its core shape. Each bullet below is expected to become its **own GitHub issue** when picked up —
this file is the backlog/grouping, not a single unit of work.

> **Revised after #194 landed:** the push channel (section A) has been extracted to #238; section B was corrected —
> the generic rule abstraction was deliberately deferred by #193 and is a **prerequisite** here, not an existing
> artifact; sections E and F capture two deferred items that `docs/domain/ALERTS.md` already scopes to this umbrella.

### A. Real notification channels
- **Push → extracted to #238** (FCM, backend half — the Android half is owned by the UI project). #238 also fixes
  the channel-composition semantics every further channel inherits: a composite dispatcher where the persisted
  channel stays authoritative (drives `last_notified_at`) and real channels are best-effort.
- Email and Telegram remain candidates (one issue each). All plug into the existing `NotificationDispatcher` port
  from #194 — no core changes. Each channel gated by an explicit `enabled` config flag.

### B. New alert rule types
- `DEVICE_DETACHED` — a device stuck in `DETACHED` for longer than a threshold (see `docs/domain/DEVICES.md`).
- `SENSOR_STALE` — a sensor whose current readings stopped updating for longer than a threshold.
- ❗️ **Prerequisite: the generic rule abstraction does not exist yet.** #193 deliberately deferred it (no generic
  interface while there is a single rule): the first new rule starts by extracting an `AlertRule` abstraction from
  `BatteryLowRule` + `EvaluateAlertsService` (today hardcoded to the battery case) — see section E. Same state
  machine (`evaluateAlert`), different rules. Mind the same skip-on-absent / absence-is-not-resolution guard.

### C. Per-target threshold & cadence overrides
- Per-device (or per-area/provider) overrides for `BATTERY_LOW` thresholds and for the reminder cadence, layered on
  top of the global defaults introduced in #193 and #194.

### D. Manual acknowledge / snooze
- Let a user acknowledge an open alert (suppress reminders while still open) or snooze it for a period.
- Requires extending the alert lifecycle/state and the reminder scheduler to honour ack/snooze.
[...]

PR #228, opened 2026-07-04
issue #193: open and resolve BATTERY_LOW alerts automatically on the …

### LINKED ISSUE #193: Alert evaluation engine + `BATTERY_LOW` rule
# Alert evaluation engine + `BATTERY_LOW` rule

## Context
With the battery level in cache (#191) and the Alert aggregate + store (#192) in place, this issue wires them together:
an **evaluation engine** that periodically checks active rules against fresh data and drives the alert lifecycle.
It ships the first concrete rule, `BATTERY_LOW`, and is the first point where the feature works end-to-end
(low battery → an alert automatically appears in `GET /alerts`).

## Decision
- **Evaluation runs on the success path of the polling cycle**, exactly like the Home Stream publish
  (`HomeStateRefreshPublisher.publish()` at the end of `FetchSensorReadingsService`). Rationale (mirrors
  `docs/domain/HOME_STREAM.md`): "readings updated ⇒ conditions may have changed" is application logic; keeping it
  on the success path means **a failed polling cycle produces no evaluation and therefore no false alerts** on
  stale data.
- **Thresholds are global** for now (a single default `trigger`/`clear` for `BATTERY_LOW`). Per-device overrides
  are deferred to a later issue.
- **Absence of a cached value ⇒ "not evaluable" ⇒ skip**, never "condition cleared". This is the critical
  consequence of the cache+DB asymmetry: Redis gives no presence guarantee (cold start, eviction, TTL), so a
  missing battery value must **not** resolve an open alert (it would close → reopen → re-notify in a loop). Same
  principle already adopted in `docs/domain/DEVICES.md` (absence of data ≠ condition ceased). State transitions
  happen **only on a fresh, confirmed reading**. Consequence: an alert (DB) can outlive the value that created it
  (cache), and stays `OPEN` until a fresh reading proves the value crossed the `clear` threshold.
- **`DETACHED` devices: leave battery alerts untouched.** A device going `DETACHED` stops refreshing its readings,
  so skip-on-absent already **freezes** its battery alert in place — no resolve, no churn. We deliberately do
  **not** resolve-and-re-open on offline blips (that would produce the exact resolve/open notification flapping the
  hysteresis exists to prevent). "Device offline" is a distinct condition, owned by the future `DEVICE_DETACHED`
  rule (#195) — not conflated with battery.
- **Battery is numeric (`Int` %, SwitchBot).** `BATTERY_LOW` compares against numeric `trigger`/`clear` thresholds.
  Netatmo's categorical `battery_state` and any cross-provider normalization are deferred (see #191).

## Scope
- Generic `AlertRule` abstraction:
  - `type`, `scope` (device/provider/global), `params` (e.g. `triggerThreshold`, `clearThreshold`).
  - Designed so future rules (`DEVICE_DETACHED`, `SENSOR_STALE`) plug into the same engine.
- One concrete rule: `BATTERY_LOW` (configurable global `trigger`/`clear`, e.g. trigger ≤ 15%, clear ≥ 25%).
- `AlertEvaluationService` (core):
  - reads the current battery value(s) from cache (#191);
  - evaluates active rules, diffs against persisted `OPEN` alerts (#192), and produces the transitions
    (open / keep / resolve) through the alert repository;
  - applies the **skip-on-absent** guard.
- Hook the evaluation into the polling success path (next to / after the existing readings refresh).

## Out of scope
- Notifications / delivery / reminder cadence (#194) — this issue only mutates alert *state*.
- Per-device threshold overrides (later issue).
- Additional rule types beyond `BATTERY_LOW` (#195+).

## Technical notes
- Configure global thresholds via Spring properties (e.g. `alerts.battery-low.trigger`,
  `alerts.battery-low.clear`), consistent with how `home.stream.*` is configured.
- Keep the engine a pure consumer of state: it reads battery from the cache port (#191) and alerts from the
  repository (#192); it must not call device drivers directly.
- The polling cadence is inherited from `DevicesDataFetchScheduler` (every minute); evaluation reuses it — **no new
  timer**.
- ❗️ **Ordering: evaluate before publishing the home state.** `FetchSensorReadingsService` already calls
  `homeStateRefreshPublisher.publish()` at the end of its success path. Alert evaluation must run **before** that
  `publish()`, so the dashboard pushed over SSE in the same cycle already reflects the new/resolved alerts (the
  projection lands in #196). This means **no new SSE trigger is needed** for polling-driven alert changes — they ride
  the existing publish.
- **Invariant for future work:** any alert state transition that can occur **outside** the polling cycle must
  publish a `HomeStateRefresh` itself (e.g. manual ack/snooze in #195), the same way config-change use cases already
  do. Within this issue all transitions are polling-driven, so the ordering above is sufficient.
- Follow TDD (`docs/TDD_WORKFLOW.md`): unit tests for the engine (open below trigger, keep while between thresholds,
  resolve above clear, **skip when value absent**, idempotency), plus an integration test for the
  polling → alert flow. Add a regression guard test on the skip-on-absent behaviour.

## Acceptance criteria
- A battery reading below `trigger` opens exactly one `BATTERY_LOW` alert, visible in `GET /alerts`.
- A subsequent reading still below `clear` keeps the same alert (no duplicate).
- A reading at/above `clear` resolves the alert.
- A missing cached battery value leaves the alert state **unchanged** (no resolve, no open).
- A failed polling cycle triggers no evaluation.

## Dependencies
- Requires #191 (battery in cache) and #192 (alert domain + store).

## Definition of done
- `CHANGELOG.md`: add an entry under `## [Unreleased]` (`Added` — automatic `BATTERY_LOW` alerts).
- Update `docs/domain/ALERTS.md` with the evaluation flow, the polling hook, global thresholds, and the
  skip-on-absent rule. Cross-link `HOME_STREAM.md` and `DEVICES.md`.

AGRFesta:
…polling cycle

### DIFF
--- a/core/src/main/kotlin/org/agrfesta/sh/api/core/application/ports/inbounds/EvaluateAlertsUseCase.kt
+++ b/core/src/main/kotlin/org/agrfesta/sh/api/core/application/ports/inbounds/EvaluateAlertsUseCase.kt
@@ -1,0 +1,18 @@
+package org.agrfesta.sh.api.core.application.ports.inbounds
+
+import org.agrfesta.sh.api.core.domain.devices.Device
+
+interface EvaluateAlertsUseCase {
+
+    /**
+     * Evaluates the active alert rules against the freshest cached data for the given [devices],
+     * driving the alert lifecycle: it opens, keeps or resolves persisted alerts accordingly.
+     *
+     * Best-effort by design (it runs on the polling success path): a device whose data is absent or
+     * unreadable is skipped — absence of data never resolves an alert — and per-device persistence
+     * failures are logged without stopping the remaining evaluations.
+     *
+     * @param devices the devices to evaluate, as fetched by the current polling cycle.
+     */
+    fun execute(devices: Collection<Device>)
+}
--- a/core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/EvaluateAlertsService.kt
+++ b/core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/EvaluateAlertsService.kt
@@ -1,0 +1,65 @@
+package org.agrfesta.sh.api.core.application.usecases
+
+import arrow.core.getOrElse
+import org.agrfesta.sh.api.core.application.ports.inbounds.EvaluateAlertsUseCase
+import org.agrfesta.sh.api.core.application.ports.outbounds.RandomGenerator
+import org.agrfesta.sh.api.core.application.ports.outbounds.TimeProvider
+import org.agrfesta.sh.api.core.application.ports.outbounds.alerts.AlertsRepository
+import org.agrfesta.sh.api.core.application.ports.outbounds.devices.DeviceBatteryRepository
+import org.agrfesta.sh.api.core.domain.alerts.Alert
+import org.agrfesta.sh.api.core.domain.alerts.AlertStatus
+import org.agrfesta.sh.api.core.domain.alerts.AlertSubject
+import org.agrfesta.sh.api.core.domain.alerts.AlertTarget
+import org.agrfesta.sh.api.core.domain.alerts.AlertTransition
+import org.agrfesta.sh.api.core.domain.alerts.AlertType
+import org.agrfesta.sh.api.core.domain.alerts.BatteryLowRule
+import org.agrfesta.sh.api.core.domain.alerts.evaluateAlert
+import org.agrfesta.sh.api.core.domain.devices.Device
+import org.agrfesta.sh.api.utils.LoggerDelegate
+import org.springframework.stereotype.Service
+
+@Service
+class EvaluateAlertsService(
+    private val alertsRepository: AlertsRepository,
+    private val deviceBatteryRepository: DeviceBatteryRepository,
+    private val batteryLowRule: BatteryLowRule,
+    private val randomGenerator: RandomGenerator,
+    private val timeProvider: TimeProvider
+) : EvaluateAlertsUseCase {
+
+    private val logger by LoggerDelegate()
+
+    override fun execute(devices: Collection<Device>) {
+        val openAlerts = alertsRepository.getAlerts(AlertStatus.OPEN)
+            .onLeft { failure ->
+                logger.warn("Skipping alert evaluation: unable to read the open alerts: $failure")
+            }
+            .getOrElse { return }
+        devices.forEach { device -> evaluateBatteryLow(device, openAlerts) }
+    }
+
+    private fun evaluateBatteryLow(device: Device, openAlerts: Collection<Alert>) {
+        val level = deviceBatteryRepository.findBy(device)
+            .onLeft { failure ->
+                logger.warn("Failed to read cached battery for device ${device.uuid}: $failure")
+            }
+            .getOrElse { return }
+            ?: return // skip-on-absent: no cached value ⇒ not evaluable, never "condition cleared"
+        val target = AlertTarget.Device(device.uuid)
+        val current = openAlerts.firstOrNull { it.type == AlertType.BATTERY_LOW && it.target == target }
+        val transition = evaluateAlert(
+            current = current,
+            conditionMet = batteryLowRule.conditionMet(level = level, alertOpen = current != null),
+            subject = AlertSubject(type = AlertType.BATTERY_LOW, target = target, details = "battery=$level%"),
+            newId = randomGenerator.uuid(),
+            at = timeProvider.now()
+        )
+        when (transition) {
+            is AlertTransition.Opened -> alertsRepository.create(transition.alert)
+                .onLeft { failure -> logger.warn("Failed to open alert for device ${device.uuid}: $failure") }
+            is AlertTransition.Resolved -> alertsRepository.resolve(transition.alert)
+                .onLeft { failure -> logger.warn("Failed to resolve alert '${transition.alert.uuid}': $failure") }
+            AlertTransition.Unchanged -> Unit
+        }
+    }
+}
--- a/persistence/src/main/kotlin/org/agrfesta/sh/api/persistence/jdbc/adapters/AlertsJdbcAdapter.kt
+++ b/persistence/src/main/kotlin/org/agrfesta/sh/api/persistence/jdbc/adapters/AlertsJdbcAdapter.kt
@@ -9,6 +9,7 @@ import org.agrfesta.sh.api.core.domain.alerts.AlertStatus
 import org.agrfesta.sh.api.core.domain.failures.AlertAlreadyOpen
 import org.agrfesta.sh.api.core.domain.failures.AlertCreationFailure
 import org.agrfesta.sh.api.core.domain.failures.AlertRepositoryError
+import org.agrfesta.sh.api.core.domain.failures.AlertResolutionFailure
 import org.agrfesta.sh.api.core.domain.failures.GetAlertsFailure
 import org.agrfesta.sh.api.persistence.jdbc.repositories.AlertsJdbcRepository
 import org.agrfesta.sh.api.utils.LoggerDelegate
@@ -43,6 +44,18 @@ class AlertsJdbcAdapter(
         AlertAlreadyOpen.left()
     } catch (e: DataAccessException) {
         logger.error("Unexpected persistence error creating alert '${alert.uuid}'", e)
+        AlertRepositoryError.left()
+    }
+
+    override fun resolve(alert: Alert): Either<AlertResolutionFailure, Unit> = try {
+        if (alertsRepo.updateResolution(alert) == 0) {
+            logger.error("Unable to resolve alert '${alert.uuid}': no such row")
+            AlertRepositoryError.left()
+        } else {
+            Unit.right()
+        }
+    } catch (e: DataAccessException) {
+        logger.error("Unexpected persistence error resolving alert '${alert.uuid}'", e)
         AlertRepositoryError.left()
     }
 }
--- a/persistence/src/main/kotlin/org/agrfesta/sh/api/persistence/jdbc/repositories/AlertsJdbcRepository.kt
+++ b/persistence/src/main/kotlin/org/agrfesta/sh/api/persistence/jdbc/repositories/AlertsJdbcRepository.kt
@@ -46,6 +46,22 @@ class AlertsJdbcRepository(
         )
         jdbcTemplate.update(sql, params)
     }
+
+    /** Updates the lifecycle columns of an existing alert row; returns the number of affected rows. */
+    fun updateResolution(alert: Alert): Int {
+        val sql = """
+            UPDATE smart_home.alert
+            SET status = :status, resolved_at = :resolvedAt
+            WHERE uuid = :uuid
+        """
+        val resolvedAt = (alert.lifecycle as? AlertLifecycle.Resolved)?.resolvedAt
+        val params = mapOf(
+            "uuid" to alert.uuid,
+            "status" to alert.lifecycle.status.name,
+            "resolvedAt" to resolvedAt?.let { Timestamp.from(it) }
+        )
+        return jdbcTemplate.update(sql, params)
+    }
 }
 
 object AlertRowMapper : RowMapper<AlertEntity> {
--- a/providers/src/testFixtures/kotlin/org/agrfesta/sh/api/providers/switchbot/SwitchBotClientAsserter.kt
+++ b/providers/src/testFixtures/kotlin/org/agrfesta/sh/api/providers/switchbot/SwitchBotClientAsserter.kt
@@ -13,11 +13,18 @@ class SwitchBotClientAsserter(
     private val mapper: ObjectMapper
 ) {
 
-    fun givenSensorData(sensorProviderId: String, data: ThermoHygroData) {
+    // battery defaults to a deterministic full charge: a random default would make any test that
+    // triggers a polling cycle nondeterministically cross the BATTERY_LOW hysteresis band.
+    fun givenSensorData(
+        sensorProviderId: String,
+        data: ThermoHygroData,
+        battery: Int = 100
+    ) {
         coEvery { switchBotDevicesClient.getDeviceStatus(sensorProviderId) } returns
                 mapper.aSwitchBotDeviceStatusResponse(
                     humidity = data.relativeHumidity.value.movePointRight(2).toInt(),
-                    temperature = data.temperature
+                    temperature = data.temperature,
+                    battery = battery
                 )
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Open and resolve `BATTERY_LOW` alerts automatically on the polling cycle, driven by global hysteresis
+  thresholds (`alerts.battery-low.trigger`/`clear`); low-battery devices now surface in `GET /alerts`
+  without manual intervention. (#193)
+
 ### Changed
 - Return a typed failure or a per-heater outcome report from the heating evaluation instead of
   swallowing every error in logs; the heating scheduler now logs an error when the evaluation fails. (#201)
--- a/app/src/main/kotlin/org/agrfesta/sh/api/configuration/AlertsConfiguration.kt
+++ b/app/src/main/kotlin/org/agrfesta/sh/api/configuration/AlertsConfiguration.kt
@@ -1,0 +1,20 @@
+package org.agrfesta.sh.api.configuration
+
+import org.agrfesta.sh.api.core.domain.alerts.BatteryLowRule
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+
+@Configuration
+class AlertsConfiguration {
+
+    /**
+     * Global `BATTERY_LOW` thresholds (per-device overrides are a future issue). The rule itself
+     * rejects a band where `clear <= trigger` at startup.
+     */
+    @Bean
+    fun batteryLowRule(
+        @Value("\${alerts.battery-low.trigger:15}") trigger: Int,
+        @Value("\${alerts.battery-low.clear:25}") clear: Int
+    ) = BatteryLowRule(trigger = trigger, clear = clear)
+}
--- a/app/src/main/resources/application.properties
+++ b/app/src/main/resources/application.properties
@@ -2,6 +2,10 @@ spring.application.name=smart-home-api
 # TTL for the per-device battery cache key (devices:{provider}:{id}:battery).
 # A small multiple of the polling interval so an offline device lets its value expire (#191).
 smart-home.cache.battery-ttl=5m
+# Global BATTERY_LOW thresholds (%): the alert opens at trigger or below and resolves only at
+# clear or above (hysteresis band, #193). clear must be strictly greater than trigger.
+alerts.battery-low.trigger=15
+alerts.battery-low.clear=25
 management.endpoints.web.exposure.include=health,info
 management.endpoint.health.show-details=when-authorized
 management.endpoint.health.probes.enabled=true
--- a/app/src/test/kotlin/org/agrfesta/sh/api/AlertsIntegrationTest.kt
+++ b/app/src/test/kotlin/org/agrfesta/sh/api/AlertsIntegrationTest.kt
@@ -3,21 +3,39 @@ package org.agrfesta.sh.api
 import arrow.core.getOrElse
 import com.fasterxml.jackson.core.type.TypeReference
 import com.fasterxml.jackson.databind.ObjectMapper
+import io.kotest.assertions.withClue
 import io.kotest.matchers.collections.shouldContainExactly
+import io.kotest.matchers.collections.shouldHaveSize
+import io.kotest.matchers.shouldBe
 import io.restassured.RestAssured.given
 import io.restassured.http.ContentType
 import org.agrfesta.sh.api.controllers.AlertResponse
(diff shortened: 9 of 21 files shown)


core/src/main/kotlin/org/agrfesta/sh/api/core/application/ports/inbounds/EvaluateAlertsUseCase.kt
```diff
@@ -0,0 +1,18 @@
...
+     * Evaluates the active alert rules against the freshest cached data for the given [devices],
+     * driving the alert lifecycle: it opens, keeps or resolves persisted alerts accordingly.
+     *
+     * Best-effort by design (it runs on the polling success path): a device whose data is absent or
+     * unreadable is skipped — absence of data never resolves an alert — and per-device persistence
+     * failures are logged without stopping the remaining evaluations.
+     *
+     * @param devices the devices to evaluate, as fetched by the current polling cycle.
+     */
+    fun execute(devices: Collection<Device>)
```
AGRFesta, line 17: 🤖 **Reviewer** — _severity: medium · confidence: medium_

`execute()` returns `Unit` and the service downgrades every failure (open-alerts read, per-device lookups, create/resolve persistence) to a log line. `docs/ARCHITECTURE.md` is explicit that inbound ports "must return functional types (`Either`)", and #201 — merged immediately before this PR — removed exactly this swallow-in-logs pattern from `EvaluateHeatingStateUseCase` in favour of a typed failure / per-heater outcome report.

The best-effort intent is legitimate and well documented in the KDoc, but best-effort and a functional contract aren't in conflict: returning an outcome report (e.g. skipped devices, failed persists, or an `Either<EvaluateAlertsFailure, AlertsEvaluationReport>` mirroring the heating design) preserves the semantics while letting the caller decide what to log. As written, `FetchSensorReadingsService` cannot even tell that the whole evaluation was skipped because the open alerts could not be read.

If you consciously chose `Unit` to keep #193 small (with the report deferred alongside the rule abstraction in #195), that's a defensible call — but then it deserves a sentence in `ALERTS.md`, because right now the code contradicts both the architecture doc and the freshest precedent in the repo.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/AlertsEvaluationService.kt
```diff
@@ -0,0 +1,65 @@
...
+import org.agrfesta.sh.api.core.domain.alerts.AlertTransition
+import org.agrfesta.sh.api.core.domain.alerts.AlertType
+import org.agrfesta.sh.api.core.domain.alerts.BatteryLowRule
+import org.agrfesta.sh.api.core.domain.alerts.evaluateAlert
+import org.agrfesta.sh.api.core.domain.devices.Device
+import org.agrfesta.sh.api.utils.LoggerDelegate
+import org.springframework.stereotype.Service
+
+@Service
+class AlertsEvaluationService(
```
AGRFesta: 🤖 **Reviewer** — _severity: low · confidence: high_

Naming convention: all 24 existing use case implementations follow `<UseCaseName>Service` (`EvaluateHeatingStateUseCase` → `EvaluateHeatingStateService`, `FetchSensorReadingsUseCase` → `FetchSensorReadingsService`, ...). `EvaluateAlertsUseCase` should therefore be implemented by `EvaluateAlertsService`, not `AlertsEvaluationService`. Pure rename, no behaviour change — but worth doing now before #195 builds the rule abstraction on top of this class name.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/EvaluateAlertsService.kt
```diff
@@ -0,0 +1,65 @@
...
+                logger.error("Failed to read cached battery for device ${device.uuid}: $failure")
+            }
+            .getOrElse { return }
+            ?: return // skip-on-absent: no cached value ⇒ not evaluable, never "condition cleared"
+        val target = AlertTarget.Device(device.uuid)
+        val current = openAlerts.firstOrNull { it.type == AlertType.BATTERY_LOW && it.target == target }
+        val transition = evaluateAlert(
+            current = current,
+            conditionMet = batteryLowRule.conditionMet(level = level, alertOpen = current != null),
+            subject = AlertSubject(type = AlertType.BATTERY_LOW, target = target, details = "battery=$level%"),
```
AGRFesta, line 53: 🤖 **Reviewer** — _severity: low · confidence: medium_

The subject is rebuilt every cycle with the fresh level (`details = "battery=$level%"`), but on `Unchanged` nothing is persisted — so `details` permanently reflects the *opening* value. A device that opened at 10% and drained to 3% still shows `battery=10%` in `GET /alerts` for its entire open lifetime, which reads as current state (the API doc example shows exactly this field).

If "details = snapshot at open time" is the intended semantics, one sentence in `ALERTS.md` would close the ambiguity; if not, the `Unchanged` branch needs an update path. Given `AlertTransition.Unchanged` is deliberately a no-op ("storage must stay unchanged"), I suspect this is intentional — flagging so it's a decision, not an accident.

providers/src/testFixtures/kotlin/org/agrfesta/sh/api/providers/switchbot/SwitchBotClientAsserter.kt
```diff
@@ -13,11 +14,16 @@ class SwitchBotClientAsserter(
     private val mapper: ObjectMapper
 ) {
 
-    fun givenSensorData(sensorProviderId: String, data: ThermoHygroData) {
+    fun givenSensorData(
+        sensorProviderId: String,
+        data: ThermoHygroData,
+        battery: Int = aRandomIntPercentage()
```
AGRFesta: 🤖 **Reviewer** — _severity: low · confidence: medium_

Defaulting `battery` to `aRandomIntPercentage()` makes every integration test that triggers a polling cycle nondeterministically cross the alert engine: with `trigger=15`, roughly 1 run in 6 of `SnapshotSensorHistoryServiceIntegrationTest` now also opens a real `BATTERY_LOW` alert in the DB as a side effect. Nothing asserts on it today (and `@CleanSmartHomeDatabase` contains the damage), but tests that exercise different code paths depending on a random draw are a latent flakiness source for any future test that observes alerts or the SSE snapshot.

A deterministic safe default (e.g. `battery: Int = 100`) keeps existing call sites out of the hysteresis band; tests that *want* a low battery pass it explicitly, as the new alerts test already does.

persistence/src/main/kotlin/org/agrfesta/sh/api/persistence/jdbc/adapters/AlertsJdbcAdapter.kt
```diff
@@ -45,4 +46,16 @@ class AlertsJdbcAdapter(
         logger.error("Unexpected persistence error creating alert '${alert.uuid}'", e)
         AlertRepositoryError.left()
     }
+
+    override fun resolve(alert: Alert): Either<AlertResolutionFailure, Unit> = try {
```
AGRFesta, line 50: 🤖 **Reviewer** — _severity: low · confidence: low_

`resolve()` persists whatever lifecycle the alert carries: called with an `Open` alert it would happily write `status = OPEN, resolved_at = NULL` and return `Right(Unit)` — a successful "resolution" that resolved nothing. The precondition ("its lifecycle must be `Resolved`") lives only in the port KDoc.

Cheap hardening options: guard in the adapter (non-`Resolved` → typed failure), or make the contract unrepresentable by having the port take the resolved parts explicitly (e.g. `resolve(uuid: UUID, resolvedAt: Instant)`), which would also let `updateResolution` hardcode `status = 'RESOLVED'` instead of trusting the caller. Low priority — the only current caller comes straight from `AlertTransition.Resolved` — but the port is public surface for future callers (manual ack in the invariant note of ALERTS.md).

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/EvaluateAlertsService.kt
```diff
@@ -0,0 +1,65 @@
...
+
+    private val logger by LoggerDelegate()
+
+    override fun execute(devices: Collection<Device>) {
+        val openAlerts = alertsRepository.getAlerts(AlertStatus.OPEN)
+            .onLeft { failure ->
+                logger.error("Skipping alert evaluation: unable to read the open alerts: $failure")
+            }
+            .getOrElse { return }
+        devices.forEach { device -> evaluateBatteryLow(device, openAlerts) }
```
AGRFesta, line 38: 🤖 **Reviewer** — _severity: low · confidence: low_

`evaluateBatteryLow` runs for *every* fetched device, including ones that cannot have a battery (mains-powered actuators), costing one Redis lookup per device per polling cycle. Correctness is safe — skip-on-absent turns them into no-ops — so this is only a minor efficiency/intent note. If the device model can already answer "battery-powered?" (the role catalogue from #222), filtering upfront would make the loop state its intent; if that information only exists at driver level, leave it as is and ignore this comment.

AGRFesta:
🤖 **Reviewer** — review summary (round 1, sha `aa32e8618d2499291525657bfae9dce704e8620c`)

Round 1 review of the BATTERY_LOW evaluation engine (#193). Solid work overall: the hysteresis rule is cleanly isolated in the domain, the skip-on-absent semantics are explicitly documented and pinned by a regression guard, the PBT is a genuine one-directional contract (no tautology), and CHANGELOG + domain docs are updated. Findings below are mostly design/convention level; the one I'd weigh most is the `Unit`-returning inbound port that swallows every failure in logs — the exact pattern #201 just removed from the heating path.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/ports/inbounds/EvaluateAlertsUseCase.kt
AGRFesta: 🧑‍💻 **Author** — Agreed on the principle; the deferral is now explicit. `Unit` was a conscious choice to keep #193 small: every failure here is a tolerated degradation on the polling success path and the caller has no meaningful branch to take today. The typed outcome report (mirroring the #201 heating design) lands in #195 together with the rule abstraction, so it is designed once for N rules. As suggested, the decision is now documented in ALERTS.md ("Evaluation" section).

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/AlertsEvaluationService.kt
AGRFesta: 🧑‍💻 **Author** — Correct, my miss on the convention. Renamed to `EvaluateAlertsService` (class + test, `git mv`), ALERTS.md reference updated.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/EvaluateAlertsService.kt
AGRFesta: 🧑‍💻 **Author** — Taking the pre-authorized exit: the model catalogue cannot answer "battery-powered?" — `DeviceFeature` is `{SENSOR, ACTUATOR}` only; battery capability exists solely at driver level (`BatteryPowered` on the driver instance). Filtering upfront would mean either extending the catalogue (out of scope) or instantiating drivers inside the evaluator, which must stay a pure consumer of state. Skip-on-absent keeps the loop correct and one Redis GET per device per minute is negligible at this scale.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/EvaluateAlertsService.kt
AGRFesta: 🧑‍💻 **Author** — Intentional (`Unchanged` must not touch storage — refreshing details would mean one UPDATE per open alert per minute), but you're right it deserved to be a documented decision: ALERTS.md now states "details is a snapshot at open time, not current state", and the API doc carries the same clarification on the `details` field.

persistence/src/main/kotlin/org/agrfesta/sh/api/persistence/jdbc/adapters/AlertsJdbcAdapter.kt
AGRFesta: 🧑‍💻 **Author** — Valid hardening, wrong moment. The only current caller builds the argument from `AlertTransition.Resolved`, so the invalid call is unreachable today; the port gains its second caller exactly in #195 (manual ack/snooze), the natural point to make the contract unrepresentable — `resolve(uuid: UUID, resolvedAt: Instant)` with `status='RESOLVED'` hardcoded in SQL — with that caller's needs on the table. Deferred to #195 and tracked in the project notes.

providers/src/testFixtures/kotlin/org/agrfesta/sh/api/providers/switchbot/SwitchBotClientAsserter.kt
AGRFesta: 🧑‍💻 **Author** — Good catch: ~1 run in 6 would silently cross the hysteresis band. Default is now a deterministic `battery: Int = 100` with a comment explaining why; low-battery tests pass the value explicitly.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/EvaluateAlertsService.kt
```diff
@@ -0,0 +1,65 @@
...
+    private val timeProvider: TimeProvider
+) : EvaluateAlertsUseCase {
+
+    private val logger by LoggerDelegate()
+
+    override fun execute(devices: Collection<Device>) {
+        val openAlerts = alertsRepository.getAlerts(AlertStatus.OPEN)
+            .onLeft { failure ->
+                logger.error("Skipping alert evaluation: unable to read the open alerts: $failure")
+            }
```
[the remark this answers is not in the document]
AGRFesta: 🧑‍💻 **Author** — Done: aligned with the #201 heating precedent — `warn` for tolerated degradations in the service, `error` stays at the persistence adapter boundary.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/EvaluateAlertsService.kt
```diff
@@ -0,0 +1,65 @@
...
+            .getOrElse { return }
+        devices.forEach { device -> evaluateBatteryLow(device, openAlerts) }
+    }
+
+    private fun evaluateBatteryLow(device: Device, openAlerts: Collection<Alert>) {
+        val level = deviceBatteryRepository.findBy(device)
+            .onLeft { failure ->
+                logger.error("Failed to read cached battery for device ${device.uuid}: $failure")
+            }
```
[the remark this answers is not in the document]
AGRFesta: 🧑‍💻 **Author** — Done: aligned with the #201 heating precedent — `warn` for tolerated degradations in the service, `error` stays at the persistence adapter boundary.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/AlertsEvaluationService.kt
```diff
@@ -0,0 +1,65 @@
...
+            conditionMet = batteryLowRule.conditionMet(level = level, alertOpen = current != null),
+            subject = AlertSubject(type = AlertType.BATTERY_LOW, target = target, details = "battery=$level%"),
+            newId = randomGenerator.uuid(),
+            at = timeProvider.now()
+        )
+        when (transition) {
+            is AlertTransition.Opened -> alertsRepository.create(transition.alert)
+                .onLeft { failure -> logger.error("Failed to open alert for device ${device.uuid}: $failure") }
+            is AlertTransition.Resolved -> alertsRepository.resolve(transition.alert)
+                .onLeft { failure -> logger.error("Failed to resolve alert '${transition.alert.uuid}': $failure") }
```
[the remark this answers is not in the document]
AGRFesta: 🧑‍💻 **Author** — Done: aligned with the #201 heating precedent — `warn` for tolerated degradations in the service, `error` stays at the persistence adapter boundary.

persistence/src/main/kotlin/org/agrfesta/sh/api/persistence/jdbc/repositories/AlertsJdbcRepository.kt
```diff
@@ -46,6 +46,22 @@ class AlertsJdbcRepository(
...
     }
+
+    /** Updates the lifecycle columns of an existing alert row; returns the number of affected rows. */
+    fun updateResolution(alert: Alert): Int {
+        val sql = """
+            UPDATE smart_home.alert
+            SET status = :status, resolved_at = :resolvedAt
+            WHERE uuid = :uuid
+        """
+        val resolvedAt = (alert.lifecycle as? AlertLifecycle.Resolved)?.resolvedAt
```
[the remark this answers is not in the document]
AGRFesta: 🧑‍💻 **Author** — Consistency call: `persist()` in this same file uses the identical `as?` pattern (merged in #192), so switching only this occurrence would make the file internally inconsistent. The line is also slated to disappear in #195, where `resolve` becomes `resolve(uuid, resolvedAt)` and the SQL hardcodes `status='RESOLVED'`. Keeping as-is for now.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/ports/inbounds/EvaluateAlertsUseCase.kt
AGRFesta: 🤖 **Reviewer**

Verified: ALERTS.md now states the `Unit` return is a conscious, #193-scoped exception to the `Either`-port rule, with the typed outcome report explicitly deferred to #195 alongside the rule abstraction — exactly the documentation trade I asked for. The deferral is a defensible call now that it's a recorded decision rather than a silent contradiction. Resolved; I'll expect the report contract when #195 lands.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/AlertsEvaluationService.kt
AGRFesta: 🤖 **Reviewer**

Verified: class and test renamed to `EvaluateAlertsService` / `EvaluateAlertsServiceTest` (git rename, 86%/98% similarity), ALERTS.md reference updated, and a repo-wide grep shows no stale `AlertsEvaluationService` references. Convention restored. Resolved.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/EvaluateAlertsService.kt
AGRFesta: 🤖 **Reviewer**

Concede — I verified the claim: `DeviceFeature` is `{SENSOR, ACTUATOR}` only, and battery capability exists solely on the driver-level `BatteryPowered` port. Filtering upfront would indeed require either widening the catalogue or pulling drivers into the evaluator, both worse than one no-op Redis GET per device per minute. This was the pre-authorized exit in my original comment. Resolved.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/EvaluateAlertsService.kt
AGRFesta: 🤖 **Reviewer**

Verified: ALERTS.md now states `details` is a snapshot at open time (with the `Unchanged`-never-touches-storage rationale), and `docs/api/alerts.md` carries the same clarification on the field consumers actually read. The write-amplification argument for not refreshing is sound. What I wanted was a decision on record, and now there is one. Resolved.

persistence/src/main/kotlin/org/agrfesta/sh/api/persistence/jdbc/adapters/AlertsJdbcAdapter.kt
AGRFesta: 🤖 **Reviewer**

Concede — deferring to #195 is the right sequencing: the only current caller builds its argument from `AlertTransition.Resolved`, so the invalid call is unreachable today, and `resolve(uuid, resolvedAt)` with `status='RESOLVED'` hardcoded is best designed when the manual-ack caller's needs are on the table. Noting for the #195 review: I'll expect the signature change there. Resolved.

providers/src/testFixtures/kotlin/org/agrfesta/sh/api/providers/switchbot/SwitchBotClientAsserter.kt
AGRFesta: 🤖 **Reviewer**

Verified: `battery: Int = 100` with a comment explaining exactly why a random default is dangerous here (nondeterministic crossing of the hysteresis band), and the low-battery integration test passes its value explicitly. The flakiness source is gone. Resolved.

AGRFesta:
🤖 **Reviewer** — review summary (round 2, sha `a377e95b4c8b51655b8bc6003d9b972d37a0e06e`)

Round 2 re-review, scoped to the delta (`aa32e86..a377e95`) and the six open reviewer threads. All four accepted findings are fixed exactly as agreed: the service is renamed to `EvaluateAlertsService` (class + test via rename, ALERTS.md updated, no stale references), the `Unit`-returning port is now a documented, #195-scoped exception in ALERTS.md, the `details`-is-a-snapshot semantics are stated in both ALERTS.md and the API doc, and the test-fixture battery default is a deterministic 100 with an explanatory comment. Both rejections were verified against the code and are factually correct (`DeviceFeature` has no battery concept; `resolve()` hardening naturally lands with the second caller in #195) — conceded. The delta introduces no new issues; all six threads close as resolved. No open findings remain from my side.

PR #229, opened 2026-07-04
issue #196: surface active alerts in GET /home and GET /devices/{uuid}

### LINKED ISSUE #196: Project alert state into existing read models
# Project alert state into existing read models

## Context
`GET /alerts` (#192) is the canonical, detailed representation of alert state. But the frontend also needs alert
awareness *inside* existing views without cross-joining endpoints: a warning icon on an area in the home dashboard,
or a "battery low" flag on a device's status. Making the FE cross-reference `/alerts` with `/home` and `/devices`
would push domain knowledge (which device belongs to which area) onto every client. The backend already owns that
knowledge (`AreasWithDevicesRepository`, used by `GetHomeDashboardService`), so it should serve these signals
pre-composed.

This issue adds **read-side projections** of alert state into the existing models. It only makes sense once #193
produces real alerts.

## Decision
- **One source of truth, multiple projections.** `GET /alerts` stays canonical. Home dashboard and device status
  expose a **minimal, derived** view computed at read time by querying the **same** `AlertRepository`. They must
  **not** re-evaluate rules nor maintain a parallel alert state.
- Keep the projection **minimal**: enough to render UX (an icon, a flag), not the full alert detail. For details
  the client follows up with `GET /alerts`.
- **No severity** (decided in #192): the warning icon derives from the **presence** of active alerts, not from a
  severity level. If severity is introduced later, the indicator can be enriched then.

## Scope
- **Home dashboard** (`GetHomeDashboardService` / `AreaDashboardDto`): add a lightweight per-area alert indicator —
  e.g. `hasActiveAlerts: Boolean`, an active-alert `count`, or the list of active alert `type`s for the area.
  Compose it by resolving the OPEN alerts whose target devices belong to each area (reuse the area→devices mapping
  already loaded by the dashboard).
- **Device status** (`GET /devices/{uuid}` — the persisted aggregate, **not** `/diagnostics`): include the device's
  active alerts, or at minimum a derived flag (e.g. `batteryLow`). Source it from the same `AlertRepository`.

## SSE delivery (home stream)
The home dashboard is also pushed over SSE (`GET /home/stream`, see `docs/domain/HOME_STREAM.md`). Once the
dashboard carries an alert indicator, that indicator must reach connected clients in real time:
- **Polling-driven alert changes ride the existing publish for free.** Alert evaluation (#193) runs on the polling
  success path **before** `homeStateRefreshPublisher.publish()`, so the dashboard broadcast in that cycle already
  includes the updated indicator. No new SSE trigger is introduced here.
- This issue's only SSE responsibility is to make the **projected indicator part of the `HomeDashboardResponse`**,
  which is exactly what the stream re-broadcasts (the stream reuses `GetHomeDashboardUseCase`).
- Alert transitions happening **outside** the polling cycle (e.g. manual ack/snooze, #195) are responsible for
  publishing their own `HomeStateRefresh` — tracked where those transitions are introduced, not here.

## Out of scope
- Changing the alert lifecycle or storage (owned by #192/#193).
- Notifications (#194).
- Adding new SSE trigger sources (none needed; see *SSE delivery* above).
- The diagnostics passthrough (`/diagnostics`) — it stays the provider's raw truth, untouched.

## Technical notes
- Inject the `AlertRepository` read port into `GetHomeDashboardService` and into the device-detail use case; compose
  exactly as the dashboard already composes readings/settings per area.
- Prefer a single batched lookup of OPEN alerts over per-device queries inside the loop, to avoid N+1 against the
  alert store (the dashboard iterates all areas/devices).
- Decide the indicator shape with the FE need in mind; keep it stable and additive to the existing response
  contracts (no breaking changes to current fields).
- Follow TDD (`docs/TDD_WORKFLOW.md`): unit tests for the dashboard/device composition (with and without active
  alerts), MVC slice + integration tests for the enriched responses.

## Acceptance criteria
- The home dashboard reports, per area, whether that area has active alerts (and, if chosen, a count / list of
  types).
- `GET /devices/{uuid}` reports the device's active alerts (or a derived flag) for a device with a low battery.
- Both are derived from the same alert store as `GET /alerts` — no duplicated evaluation, consistent results.
- Existing dashboard/device response fields are unchanged (additive only).

## Dependencies
- Requires #193 (alerts are actually produced) and #192 (alert store + read port).
- Independent of #194; can proceed in parallel with notifications.

## Definition of done
- `CHANGELOG.md`: entries under `## [Unreleased]` (`Changed` — home dashboard / device status now surface active
  alerts).
- API docs updated: `docs/api/home.md` and `docs/api/devices.md` (+ `API_INDEX.md` if needed).
- Cross-link the projection decision in `docs/domain/ALERTS.md`.

AGRFesta:


core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/GetDeviceService.kt
```diff
@@ -13,14 +17,33 @@ import java.util.UUID
...
+                )
+            }
+
+    /**
+     * Best-effort lookup of the open alerts targeting the device: a lookup failure degrades to `null`
+     * ("unknown", never a false "no alerts") but is logged so an alert store outage is not silently
+     * invisible.
+     */
+    private fun resolveActiveAlerts(deviceId: UUID): Set<AlertType>? =
+        alertsRepository.getAlerts(AlertStatus.OPEN)
```
AGRFesta, line 41: 🤖 **Reviewer** — _severity: low · confidence: medium_

**Full open-alert scan to answer a single device.** `getAlerts(AlertStatus.OPEN)` pulls every open alert and filters in memory by target. At home scale this is irrelevant and I'm fine shipping it; flagging it because the dashboard path deliberately documents its batched read as "no N+1" while this path grows linearly with total open alerts for a single-device lookup. When a second call site needs per-target alerts, add a `getAlerts(status, target)` overload to the port instead of repeating the in-memory filter. No action required now — a note for #195 is enough.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/GetHomeDashboardService.kt
```diff
@@ -67,13 +76,31 @@ class GetHomeDashboardService(
...
     }
 
+    private fun resolveActiveAlerts(
+        area: AreaDtoWithDevices,
+        openAlerts: Either<GetAlertsFailure, Collection<Alert>>
+    ): FieldResult<Set<AlertType>> =
+        openAlerts.fold(
+            { FieldFailure("Unable to retrieve active alerts") },
+            { alerts ->
+                val areaDeviceIds = (area.sensors + area.actuators).map { it.uuid }.toSet()
```
AGRFesta, line 94: 🤖 **Reviewer** — _severity: low · confidence: low_

**Blind spot: alerts on unassigned devices are invisible on the dashboard.** The projection attributes `DEVICE`-target alerts to areas via the device's assignments; `docs/domain/ALERTS.md` covers `PROVIDER`/`GLOBAL` ("not attributed to areas"), but a `BATTERY_LOW` on a paired device with **no area assignment** falls in neither bucket — it shows up only in `GET /alerts`, with no indicator anywhere on the home dashboard. That may be exactly the intended contract (the dashboard is area-centric), in which case a one-line note in ALERTS.md alongside the PROVIDER/GLOBAL sentence would close the ambiguity; if not, a future global-level indicator (e.g. on `globalState`) is the likely home for it.

AGRFesta:
🤖 **Reviewer** — review summary (round 1, sha `ee516664d140000571468eebe9b5800b897ae660`)

Round 1 review of #196 — alert projection into GET /home and GET /devices/{uuid}. (#193's evaluation engine is not part of this PR; it was merged separately via #228.) Solid PR: the projections stay minimal read-time lookups behind the existing AlertsRepository port (one source of truth, no re-derivation), the dashboard read is batched — a single getAlerts per build, pinned by a dedicated test — and lookup failures degrade to FieldFailure / null, never a false "no alerts". API docs, domain docs and changelog are all updated. Both findings below are low severity — design points, no functional bug found.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/GetDeviceService.kt
AGRFesta: 🙋 **Author** — Agreed, and it was a deliberate trade-off: at home scale the open-alert table is tiny, so the in-memory filter keeps the port unchanged (YAGNI). Tracked as a note for #195: when the rule abstraction lands, introduce a `getAlerts(status, target)` overload if a second per-target call site shows up, instead of duplicating the filter. No code change in this PR.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/GetHomeDashboardService.kt
AGRFesta: 🙋 **Author** — The area-centric contract is intended (the dashboard projects only what it can attribute; `GET /alerts` stays the canonical view), but you're right the unassigned-device case was undocumented. Added the one-line note in ALERTS.md next to the PROVIDER/GLOBAL sentence. A global-level indicator on `globalState` remains a possible future enrichment if the FE ever needs it.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/GetDeviceService.kt
```diff
@@ -13,14 +17,33 @@ import java.util.UUID
...
+            .onLeft { failure -> logger.error("Failed to read open alerts for device $deviceId: $failure") }
+            .getOrNull()
+            ?.filter { it.target == AlertTarget.Device(deviceId) }
+            ?.map { it.type }
+            ?.toSet()
```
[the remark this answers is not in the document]
AGRFesta: 🙋 **Author** — The cited "repository rule" doesn't exist in this repo (no logging-level policy in docs/ nor a .gemini styleguide). The new line deliberately mirrors its sibling `resolveBatteryLevel` in the same file, which logs the identical cache-degradation at `error`; switching only one of the two would make the file internally inconsistent. The underlying point is fair though — the codebase mixes `warn` (EvaluateAlertsService) and `error` (FetchSensorReadingsService, here) for tolerated degradations — but that's a cross-cutting policy decision, out of scope for this PR. Happy to align both helpers once a logging policy is agreed in a dedicated issue.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/GetDeviceService.kt
AGRFesta: 🤖 **Reviewer**

Resolved. The trade-off is sound at home scale and my original comment already asked for nothing more than a note for #195, which you've tracked: introduce a `getAlerts(status, target)` port overload when a second per-target call site appears, rather than duplicating the in-memory filter.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/GetDeviceService.kt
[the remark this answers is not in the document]
AGRFesta: 🤖 **Reviewer**

Conceding on the author's side — verified both claims: (1) no logging-level policy exists anywhere in `docs/` (the "repository rule" Gemini cites is not from this repo), and (2) the sibling `resolveBatteryLevel` in the same file logs the identical tolerated degradation at `error`, so changing only the new line would make the file internally inconsistent. The codebase-wide warn/error inconsistency is real but belongs to a dedicated cross-cutting issue, not this PR. Resolving.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/GetHomeDashboardService.kt
AGRFesta: 🤖 **Reviewer**

The note lands in the right place and closes the ambiguity for the dashboard, but the wording overshoots: "both surface only in `GET /alerts`" is accurate for `PROVIDER`/`GLOBAL` alerts, yet a device-target alert on an unassigned device **also** surfaces in `GET /devices/{uuid}` `activeAlerts` — the device-detail projection filters by target only, with no assignment check (see `GetDeviceService.resolveActiveAlerts`), and the very next bullet in this same section documents that projection. Suggested wording: "`PROVIDER`/`GLOBAL` alerts are not attributed to areas, and neither are alerts on devices with no current area assignment — the former surface only in `GET /alerts`, the latter in `GET /alerts` and the device detail." Reopening for this one-line fix.

AGRFesta:
🤖 **Reviewer** — review summary (round 2, sha `c50789f8ac3b822f28a5f26f3857667a1917aa8a`)

Round 2 re-review (delta ee51666..c50789f). The delta is a single doc change: the ALERTS.md note on unassigned-device alerts. Thread outcomes: the in-memory alert filter thread is resolved (accepted trade-off, follow-up tracked for #195); the Gemini logging-level thread is resolved (rejection verified: no logging policy exists in docs/, and the sibling resolveBatteryLevel logs the identical degradation at error); the unassigned-device thread is reopened for a one-word doc inaccuracy — the new note claims those alerts surface only in GET /alerts, but they also surface in GET /devices/{uuid}. No new issues in the delta.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/GetHomeDashboardService.kt
AGRFesta: 🙋 **Author** — Correct, my note overshot: the device-detail projection filters by target only (no assignment check), so an alert on an unassigned device does surface in GET /devices/{uuid} as well. Reworded as suggested: the former (PROVIDER/GLOBAL) surface only in GET /alerts, the latter also in the device detail projection documented in the next bullet.

AGRFesta: 🤖 **Reviewer**

Verified in `095f7ce`: the note now reads "the former surface only in `GET /alerts`, the latter in `GET /alerts` and the device detail projection below" — accurate on both counts, and the device-detail bullet it points to immediately follows. Resolved; this closes the last open thread of the review.

AGRFesta:
🤖 **Reviewer** — review summary (round 3, sha `095f7ce3c57fbfe5eb9fd212810abadddf781c15`)

Round 3 re-review (delta c50789f..095f7ce). The delta is exactly the requested one-line rewording of the ALERTS.md unassigned-device note. Verified against the doc: the sentence now correctly states that PROVIDER/GLOBAL alerts surface only in GET /alerts while alerts on unassigned devices also surface in the device detail projection, and the referenced projection is indeed the next bullet in the same section. The last open thread is resolved; no new issues in the delta. All threads closed — nothing outstanding for the author.

Issue #233, opened 2026-07-06
`POST /areas` silently drops `isIndoor: false` — INSERT does not persist the flag


AGRFesta:
# `POST /areas` silently drops `isIndoor: false` — INSERT does not persist the flag

**Severity:** Medium (silent data corruption on create, no error surfaced)
**Area:** `:persistence` — area write path
**Type:** Bug
**Origin:** Discovered during #231 (Area domain model extraction)

## Summary

`AreasJdbcRepository.persist` does not include the `is_indoor` column in its INSERT statement.
The column is declared `NOT NULL DEFAULT true` (migration
`V1_0_3__rename_room_in_area_and_add_is_indoor_column.sql`), so every created area is stored as
indoor regardless of the flag provided by the caller. Creating an area with `isIndoor: false`
therefore succeeds, but the persisted row has `is_indoor = true`.

Note the inconsistency within the same class: `update` **does** write `is_indoor`, so a subsequent
`PUT /areas/{id}` can "repair" the row — which makes the bug intermittent from the API consumer's
point of view.

## Evidence

- `persistence/src/main/kotlin/org/agrfesta/sh/api/persistence/jdbc/repositories/AreasJdbcRepository.kt`
  — `persist(area: Area)`: `INSERT INTO smart_home.area (uuid, name, created_on, updated_on) ...`
  — no `is_indoor` column, `area.isIndoor` is never read.
- `persistence/src/main/resources/db/migration/V1_0_3__rename_room_in_area_and_add_is_indoor_column.sql`
  — `ADD COLUMN is_indoor BOOLEAN NOT NULL DEFAULT true;` — the DB default masks the missing value.
- `CreateAreaService` returns the in-memory `Area` (with the caller's `isIndoor`) after a successful
  save, so the create flow has no chance to notice the divergence.

## Reproduction

1. `POST /areas` with body `{"name": "Garden", "isIndoor": false}` → `201 Created`.
2. `GET /areas/{id}` → response contains `"isIndoor": true`.

No existing test covers the create-with-`isIndoor: false` → read-back round trip; the whole suite
is green with the bug in place.

## Expected behavior

The `isIndoor` value provided at creation time is persisted as-is and returned by subsequent reads.

## Proposed fix

Add `is_indoor` to the INSERT column list and bind `area.isIndoor` in
`AreasJdbcRepository.persist`. This is a behavioral bug fix: follow `docs/TDD_WORKFLOW.md` —
[...]

PR #234, opened 2026-07-06
issue #231: replace AreaView with a dedicated Area domain model

### LINKED ISSUE #231: `AreaView` read-model is the de facto Area write model
# `AreaView` read-model is the de facto Area write model

**Severity:** Medium (design smell, documented deviation)
**Area:** `:core` application layer — area write path
**Type:** DDD / domain modeling
**Origin:** Follow-up of #202 (PR #230, review round 1)

## Summary

Issue #202 moved read-models out of `core/domain` into `core/application/readmodels` and renamed
them `*View`. That relocation exposed a pre-existing modeling gap: the area write path has **no
domain entity at all**. `AreasRepository.save(area: AreaView)` and `update(area: AreaView)` take the
*read-model* as input, and `CreateAreaService` / `UpdateAreaService` **construct** an `AreaView` in
order to persist it. A type named `View` that is built by command services and consumed by a write
port is actively misleading — it is the de facto Area write model.

`docs/ARCHITECTURE.md` currently records this as a **known deviation** from the read-model rule
("read-models are query shapes"); this issue tracks removing that deviation.

## Evidence

- `core/application/ports/outbounds/areas/AreasRepository.kt` — `save(area: AreaView)`,
  `update(area: AreaView)`: write operations taking the read-model as input.
- `core/application/usecases/CreateAreaService.kt` — constructs `AreaView(...)` to persist a new area.
- `core/application/usecases/UpdateAreaService.kt` — constructs `AreaView(...)` to update an area.
- `docs/ARCHITECTURE.md`, "Read-models (exception)" — the documented deviation pointing at this issue.

## Why this matters

- **Misleading semantics.** A `*View` suffix promises "query/presentation shape"; using it as the
  input to `save`/`update` erodes the naming convention introduced by #202 the moment it landed.
- **No home for Area invariants.** With no `Area` domain entity, any future business rule about
  areas (name constraints, indoor/outdoor semantics, assignment rules) has nowhere to live except
  services or the read-model itself.
- **Blocks the read-model rule from becoming absolute.** ARCHITECTURE.md cannot state
  "read-models are never input to domain logic" (and no ArchUnit rule can enforce it) while the main
  write port violates it.

## Proposed direction

1. Introduce a proper `Area` domain type in `core/domain/areas` (entity or command model —
   `uuid`, `name`, `isIndoor` today) and use it in `AreasRepository.save`/`update` and in
   `CreateAreaService` / `UpdateAreaService` / `GetAreaByIdService` (fetch-by-id arguably returns
   the entity too, since callers use it for command flows).
2. Keep `AreaView` (and `AreaWithDevicesView`) strictly on the query side
   (`GetAreasUseCase`, dashboard, heating evaluation snapshot).
3. Update persistence adapters (`AreasJdbcAdapter` & co.) to map entity ↔ row; the mapping is
   field-identical today, so this is mostly type plumbing.
4. Tighten the ARCHITECTURE.md "Read-models (exception)" section: remove the documented deviation
   and, if feasible, add an ArchUnit rule forbidding `..core.application.readmodels..` types as
   parameters of outbound-port write methods.

## Acceptance criteria

- [ ] `AreasRepository.save`/`update` accept a domain type from `core/domain/areas`, not a `*View`.
- [ ] `CreateAreaService` / `UpdateAreaService` no longer construct `AreaView`.
- [ ] `AreaView` remains used only by query paths.
- [ ] ARCHITECTURE.md no longer lists the write-path deviation.
- [ ] Existing test suite green; behavior unchanged (API contracts identical).

## Notes

Mostly type plumbing, but the new entity is a **domain model change**, not a pure move: follow
`docs/TDD_WORKFLOW.md` where behavior is touched (e.g. if invariants are added to `Area`), and rely
on the existing suite for the mechanical substitutions. Related but separate follow-up from the same
review round: ordering contract of `GET /areas/{id}/heating-schedule` intervals (sorting asymmetry
between `GetHeatingScheduleService` and `ReplaceHeatingScheduleService`).

AGRFesta:


core/src/test/kotlin/org/agrfesta/sh/api/core/CoreArchitectureTest.kt
```diff
@@ -66,5 +67,20 @@ class CoreArchitectureTest {
...
+        private val anyReadModelParameter: DescribedPredicate<List<JavaClass>> =
+            DescribedPredicate.describe("any parameter in core.application.readmodels") { parameters ->
+                parameters.any { it.packageName.contains("core.application.readmodels") }
+            }
+
+        @JvmField
+        @ArchTest
+        val readModelsAreNeverInputToOutboundPorts: ArchRule = noMethods()
+            .that().areDeclaredInClassesThat().resideInAPackage("..core.application.ports.outbounds..")
+            .should().haveRawParameterTypes(anyReadModelParameter)
```
AGRFesta: 🤖 **Reviewer** — _severity: low · confidence: high_

`haveRawParameterTypes` inspects only the **raw** parameter types, so this rule can be evaded by generics: a port method `fun saveAll(areas: Collection<AreaWithDevicesView>)` has raw parameter type `Collection` and would pass. Meanwhile ARCHITECTURE.md now states unconditionally that read-models are “never accepted as input by outbound-port methods (ArchUnit-enforced)” — the enforcement is narrower than the claim. Consider a custom condition over `JavaMethod.parameterTypes` using `JavaType.getAllInvolvedRawTypes()` (covers type arguments), or soften the doc claim. No current violation exists; this is about future-proofing the guard.

core/src/main/kotlin/org/agrfesta/sh/api/core/domain/areas/Area.kt
```diff
@@ -0,0 +1,16 @@
+package org.agrfesta.sh.api.core.domain.areas
+
+import java.util.UUID
+
+/**
+ * Domain model of an area of the home.
+ *
+ * Used by the area write path ([org.agrfesta.sh.api.core.application.ports.outbounds.areas.AreasRepository])
```
AGRFesta: 🤖 **Reviewer** — _severity: low · confidence: medium_

The KDoc of a domain model references an application-layer port (`core.application.ports.outbounds.areas.AreasRepository`). The code dependency direction is fine (docs don’t compile-link), but it inverts the documented knowledge direction — the domain should not know who consumes it — and the FQN will silently go stale if the port moves or is renamed. Suggest trimming to the second half only (“query-only shapes use the read-models in `core/application/readmodels/areas`”) or phrasing it without naming the port.

AGRFesta:
🤖 **Reviewer** — review summary (round 1, sha `d6757e998f92e3816833064a25e6d45dcfa1c496`)

Round 1 review of PR #234 (issue #231: replace AreaView with a dedicated Area domain model). Solid structural refactor: the new `Area` domain model in `core/domain/areas/` removes the documented deviation (read-model as write input), all four area use cases / `AreasRepository` / adapters are retargeted consistently, and the controller boundary gains a dedicated `AreaResponse` DTO so core types are no longer serialized directly — an improvement over the previous state. The new ArchUnit guard (`readModelsAreNeverInputToOutboundPorts`) locks the invariant in, and ARCHITECTURE.md was updated to describe the current state. HTTP contract is unchanged (same JSON shape `uuid`/`name`/`isIndoor`, verified by the untouched MVC-slice response assertions), so the absence of CHANGELOG and API-doc updates is correct per CLAUDE.md; TDD skip is legitimate for a pure structural refactor. Note (no action in this PR): with `Area` now the official write model, `AreasJdbcRepository.persist` still not inserting `is_indoor` (open issue #233) becomes a sharper domain-contract violation — worth prioritizing that follow-up. Two low-severity findings inline.

core/src/main/kotlin/org/agrfesta/sh/api/core/domain/areas/Area.kt
AGRFesta: 🤖 **Author** — accepted.

Agreed: the domain model shouldn't name its consumers, and the FQN would go stale silently.
Trimmed the KDoc to the read-model pointer only ("Query-only shapes use the read-models in
`core/application/readmodels/areas`"). Fixed in the next commit.

core/src/test/kotlin/org/agrfesta/sh/api/core/CoreArchitectureTest.kt
AGRFesta: 🤖 **Author** — accepted.

Correct: `haveRawParameterTypes` only sees raw types, so `Collection<AreaWithDevicesView>` would
slip through while ARCHITECTURE.md claims an absolute rule. Strengthened the guard instead of
softening the doc: replaced the predicate with a custom `ArchCondition` over
`JavaMethod.parameterTypes` using `getAllInvolvedRawTypes()`, which covers type arguments as well.
The ARCHITECTURE.md claim now matches the enforcement. Fixed in the next commit.

core/src/main/kotlin/org/agrfesta/sh/api/core/domain/areas/Area.kt
AGRFesta: 🤖 **Reviewer**

Verified: the KDoc now contains only the read-model pointer (`core/application/readmodels/areas`); the reference to the application-layer port is gone. The domain model no longer documents its consumers. Resolved.

core/src/test/kotlin/org/agrfesta/sh/api/core/CoreArchitectureTest.kt
AGRFesta: 🤖 **Reviewer**

Verified: the rule now uses a custom `ArchCondition` over `JavaMethod.parameterTypes` with `getAllInvolvedRawTypes()`, so read-models nested in generic parameters (e.g. `Collection<AreaWithDevicesView>`) are caught as well. The enforcement now matches the unconditional claim in ARCHITECTURE.md, and the violation message names the offending read-model types. `CoreArchitectureTest` passes on the current head. Resolved.

AGRFesta:
🤖 **Reviewer** — review summary (round 2, sha `d5d4942ee98e61f43b9f5af0aa373da00e1d62b4`)

Round 2 re-review, scoped to the delta d6757e9..d5d4942. Both round-1 findings were accepted and fixed: the Area KDoc no longer names its application-layer consumer (only the read-model pointer remains), and the ArchUnit guard was strengthened with a custom ArchCondition over parameterTypes using getAllInvolvedRawTypes(), closing the generics loophole so the enforcement now matches the unconditional claim in ARCHITECTURE.md. A follow-up commit reorders imports in three test files to satisfy detekt; no semantic change. Verified locally: CoreArchitectureTest passes and detekt is clean on the current sources. No new findings — both threads resolved, review converged.

PR #235, opened 2026-07-07
issue #233: persist the isIndoor flag in AreasJdbcRepository.persist

### LINKED ISSUE #233: `POST /areas` silently drops `isIndoor: false` — INSERT does not persist the flag

AGRFesta:


AGRFesta:
🤖 **Reviewer** — review summary (round 1, sha `ccb2786a25c373c00839e2b415e5c75005874438`)

Reviewed the full diff (origin/dev...ccb2786). Clean, well-scoped bugfix: the INSERT in `AreasJdbcRepository.persist` now includes `is_indoor`, aligning it with `update()` and the row mapper; the DB default `true` explains why the bug only surfaced for `isIndoor = false`. Test coverage pins both directions at the persistence slice level (the `false` case is the regression pin, the `true` case a guard), with explicit subject setup and descriptive clues. Changelog entry is correct under `Fixed`; no API-doc update owed since `docs/api/areas.md` already documented the intended contract. The import-ordering violation introduced in the first commit was already fixed in ccb2786. No inline findings — approving as-is. Minor process note: consider squashing the `fixup` commit before merge.

PR #236, opened 2026-07-08
issue #232: sort heating-schedule intervals by startTime in a shared …

### LINKED ISSUE #232: `GET /areas/{id}/heating-schedule` returns intervals in unspecified order
# `GET /areas/{id}/heating-schedule` returns intervals in unspecified order

**Severity:** Low (contract gap, cosmetic impact)
**Area:** `:core` application layer — heating schedule query
**Type:** API contract / consistency
**Origin:** Follow-up of #202 (PR #230, review round 1; originally flagged by Gemini Code Assist)

## Summary

`ReplaceHeatingScheduleService` sorts intervals by `startTime` before building the
`HeatingScheduleView` it returns, while `GetHeatingScheduleService` preserves whatever order the
repository yields. The ordering of `intervals` in the GET response is therefore an accident of
persistence, not a contract: the same schedule can render in different orders on write response vs
subsequent reads, causing UI display glitches. The fix was deliberately kept out of #202 because
adding sorting there would have been a behavior change inside a pure structural refactor.

## Evidence

- `core/application/usecases/GetHeatingScheduleService.kt` — maps
  `setting.temperatureSchedule` to `IntervalView`s in repository order.
- `core/application/usecases/ReplaceHeatingScheduleService.kt` — builds its response with
  `intervals.sortedBy { it.startTime }`.
- `docs/api/` heating-schedule documentation does not specify any ordering for `intervals`.

## Why this matters

- **Unstable contract.** Consumers either re-sort defensively or render intervals in
  nondeterministic order; the write response and the read response can disagree for the same data.
- **Asymmetry invites drift.** Two services shaping the same view with different rules is exactly
  the kind of divergence #202 tried to make visible by centralizing read-models.

## Proposed direction

1. Define the contract: `intervals` are always sorted by `startTime` ascending.
2. Apply the sort in `GetHeatingScheduleService` (mirroring `ReplaceHeatingScheduleService`), or —
   better — in a single shared mapping point for `HeatingScheduleView` so the rule cannot diverge
   again.
3. Add a test asserting the ordering contract on the GET path (repository returning unordered
   intervals → response sorted).
4. Document the ordering in `docs/api/<resource>.md` for the heating-schedule endpoints.

## Acceptance criteria

- [ ] `GET /areas/{id}/heating-schedule` returns intervals sorted by `startTime` ascending,
      regardless of repository order.
- [ ] A test guards the ordering contract.
- [ ] API docs state the ordering explicitly.
- [ ] `CHANGELOG.md` entry under `Changed` (observable contract clarification).

## Notes

This is a **behavior change** (small, observable): follow `docs/TDD_WORKFLOW.md` — red test asserting
sorted output first, then the fix. Sibling follow-up from the same review round:
`ISSUE-area-write-model.md` (Area write path lacks a domain entity).

AGRFesta:
…HeatingScheduleView factory

### DIFF
--- a/core/src/main/kotlin/org/agrfesta/sh/api/core/application/readmodels/areas/HeatingScheduleView.kt
+++ b/core/src/main/kotlin/org/agrfesta/sh/api/core/application/readmodels/areas/HeatingScheduleView.kt
@@ -1,12 +1,36 @@
 package org.agrfesta.sh.api.core.application.readmodels.areas
 
+import org.agrfesta.sh.api.core.domain.areas.TemperatureInterval
 import org.agrfesta.sh.api.core.domain.commons.Temperature
 import java.time.LocalTime
 
 data class HeatingScheduleView(
     val defaultTemperature: Temperature,
     val intervals: List<IntervalView>
-)
+) {
+    companion object {
+        /**
+         * Builds the view from domain intervals, sorting them by [TemperatureInterval.startTime]
+         * ascending — the ordering contract of `GET`/`PUT /areas/{areaId}/heating-schedule`.
+         *
+         * This factory is the single point applying that contract: build the view through it,
+         * not via the primary constructor, whenever intervals come from an unordered source.
+         */
+        fun from(
+            defaultTemperature: Temperature,
+            intervals: Collection<TemperatureInterval>
+        ) = HeatingScheduleView(
+            defaultTemperature = defaultTemperature,
+            intervals = intervals.sortedBy { it.startTime }.map {
+                IntervalView(
+                    temperature = it.temperature,
+                    startTime = it.startTime,
+                    endTime = it.endTime
+                )
+            }
+        )
+    }
+}
 
 data class IntervalView(
     val temperature: Temperature,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   without manual intervention. (#193)
 
 ### Changed
+- Return `intervals` sorted by `startTime` ascending on `GET /areas/{areaId}/heating-schedule`,
+  matching the PUT response; ordering was previously unspecified. (#232)
 - Return a typed failure or a per-heater outcome report from the heating evaluation instead of
   swallowing every error in logs; the heating scheduler now logs an error when the evaluation fails. (#201)
 - Surface per-area active alert types in the home dashboard as an additive `activeAlerts` field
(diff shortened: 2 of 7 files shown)


core/src/main/kotlin/org/agrfesta/sh/api/core/application/readmodels/areas/HeatingScheduleView.kt
```diff
@@ -1,12 +1,29 @@
...
 import org.agrfesta.sh.api.core.domain.commons.Temperature
 import java.time.LocalTime
 
 data class HeatingScheduleView(
     val defaultTemperature: Temperature,
     val intervals: List<IntervalView>
-)
+) {
+    companion object {
+        fun from(
```
AGRFesta, line 19: 🤖 **Reviewer** — _severity: low · confidence: high_

**Missing KDoc on the factory's sorting contract.** The entire reason `from(...)` exists is the guarantee that `intervals` come out sorted by `startTime` ascending — but nothing at the call site or signature reveals that; a reader of `GetHeatingScheduleService` now sees strictly less than before (the explicit `sortedBy` moved out of view). ARCHITECTURE.md: *"DO add KDoc to every public method ... whose contract is non-obvious"*. Suggest a short KDoc, e.g.: `Builds the view from domain intervals, sorting them by startTime ascending — the ordering contract of GET/PUT /areas/{areaId}/heating-schedule.` Minor related note: the primary constructor stays public (needed by tests/`copy`), so the guarantee holds only for callers going through `from` — the KDoc making the contract explicit is what keeps future call sites honest.

AGRFesta:
🤖 **Reviewer** — review summary (round 1, sha `a57d8dea83463b1d0ae66d53e8f0bad7ce4398aa`)

Round 1 review of #232 (sort heating-schedule intervals by startTime in a shared HeatingScheduleView factory). Solid, focused change: the duplicated mapping in Get/Replace services is centralized in HeatingScheduleView.from, both services are pinned by new unit tests (the GET one RED-drives the fix, the Replace one is a legitimate green-on-arrival guard), and CHANGELOG + docs/api/areas.md are updated for both GET and PUT sections. One low-severity finding on documenting the factory's sorting contract.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/readmodels/areas/HeatingScheduleView.kt
AGRFesta: 🤖 **Author** — decision: accepted

KDoc added to `HeatingScheduleView.from` documenting the sorting contract
(sorted by `startTime` ascending — the ordering contract of
`GET`/`PUT /areas/{areaId}/heating-schedule`) and pointing future call sites
at the factory as the single point applying it. Addressed in eb45572.

On the constructor-visibility note: agreed it stays public — slice tests build
the view directly and `copy` relies on it; the KDoc is the proportionate guard.

AGRFesta: 🤖 **Reviewer**

Verified in eb45572: the KDoc on `HeatingScheduleView.from` states the sorting contract (sorted by `startTime` ascending, the ordering contract of `GET`/`PUT /areas/{areaId}/heating-schedule`) and directs future call sites through the factory rather than the primary constructor — exactly the guard requested. Resolving.

AGRFesta:
🤖 **Reviewer** — review summary (round 2, sha `eb45572e347cfae1d11240a97244c5f14659fd43`)

Round 2 re-review scoped to the delta since a57d8de (single commit eb45572). The delta contains exactly the accepted fix: a KDoc on HeatingScheduleView.from documenting the startTime-ascending ordering contract of GET/PUT /areas/{areaId}/heating-schedule and pointing future call sites at the factory. Verified it fully addresses the round-1 finding; no new issues in the delta. All threads resolved — LGTM.

PR #237, opened 2026-07-11
issue #194: notify alert transitions with reminder cadence and retent…

### LINKED ISSUE #194: Notifications and reminder policy
# Notifications and reminder policy

## Context
Issues #192–#193 make alerts open, persist and resolve automatically. This issue adds the **notification** side: the
*delivery* projection of an alert. A notification is an ephemeral event/record produced by an alert transition —
not a domain state. The alert remains the source of truth.

This issue covers the **mechanism and cadence**, with a **minimal persisted delivery channel** only. Real channels
(push / email / Telegram) are deferred to #195.

## Decision
- Three notification events per alert:
  - `opened` — sent **immediately** the first time the alert opens;
  - `reminder` — re-sent **periodically** while the alert stays `OPEN`, at a configurable cadence
    (e.g. daily / weekly); **not** on every polling cycle;
  - `resolved` — sent when the alert resolves (**included** in this issue).
- Delivery sits **behind an outbound port** (`NotificationDispatcher`), with a minimal implementation that
  **persists** the notification (queryable via API). This mirrors how `HomeStateRefreshPublisher` abstracts its
  delivery mechanism so it can be swapped without touching the core. Real channels become #195 implementations of the
  same port.
- Reminder cadence is enforced via a `last_notified_at` field on the alert (added/reserved in #192) plus a scheduler
  that respects the configured interval.

## Scope
- `NotificationDispatcher` outbound port.
- Minimal persisted implementation: a `notifications` table (Flyway migration) + JDBC adapter recording each
  emitted notification (alert ref, event type `opened`/`reminder`/`resolved`, timestamp, payload).
- Wire dispatch into the alert transitions from #193:
  - on open → `opened`;
  - on resolve → `resolved`.
- Reminder scheduler: periodically scans `OPEN` alerts whose `last_notified_at` is older than the configured
  cadence, emits a `reminder`, and updates `last_notified_at`.
- `GET /notifications` read endpoint (or extend the alerts API) to inspect emitted notifications, **paginated**
  (the table is unbounded over time).
- **Retention policy:** a scheduled prune that deletes notifications older than a configurable window
  (e.g. `alerts.notifications.retention`, default 90 days), so the table does not grow without bound.

## Out of scope
- Real delivery channels: push / email / Telegram (#195 — separate implementations of `NotificationDispatcher`).
- New alert rule types (#195).
- Manual acknowledge / snooze (#195).
- Per-device reminder cadence (global cadence for now).

## Technical notes
- Configure cadence via Spring properties (e.g. `alerts.notifications.reminder-interval` as ISO-8601 duration,
  default daily), consistent with existing config style.
- The reminder scheduler is a thin driving adapter (`cron → use case`), like `DevicesDataFetchScheduler`; the
  re-notify decision logic lives in a core use case.
- Set `last_notified_at` on the `opened` emission so the first reminder is correctly spaced from the open event.
- Keep dispatch on the success path of the corresponding alert transition (no notification if the transition save
  fails).
- **Use the `TimeProvider` outbound port** for all timestamps and for the "reminder due" comparison — never
  `Instant.now()` — for testability, consistent with `GetHomeDashboardService`.
- **`GET /notifications` sits behind the API-key auth** (SHA-256, `docs/SECURITY.md`), like every other endpoint.
- Follow TDD (`docs/TDD_WORKFLOW.md`): unit tests for the reminder-due logic (due / not-due / just-opened),
  dispatch on open and on resolve, adapter slice test, integration test for the end-to-end notify flow.

## Acceptance criteria
- Opening a `BATTERY_LOW` alert emits exactly one `opened` notification.
- While the alert stays open, a `reminder` is emitted no more often than the configured cadence (not every minute).
- Resolving the alert emits a `resolved` notification.
- Emitted notifications are inspectable via the API (paginated, behind API-key auth).
- Notifications older than the configured retention window are pruned.

## Dependencies
- Requires #193 (alert transitions) and #192 (alert store with `last_notified_at`).

## Definition of done
- `CHANGELOG.md`: add entries under `## [Unreleased]` (`Added` — notifications + reminder cadence; new endpoint).
- API docs: `docs/api/notifications.md` (or extend `alerts.md`) + `API_INDEX.md`.
- Update `docs/domain/ALERTS.md` with the notification projection, the three events, and the cadence policy.

AGRFesta:
…ion via GET /notifications

### DIFF
--- a/app/src/main/kotlin/org/agrfesta/sh/api/schedulers/AlertRemindersScheduler.kt
+++ b/app/src/main/kotlin/org/agrfesta/sh/api/schedulers/AlertRemindersScheduler.kt
@@ -1,0 +1,27 @@
+package org.agrfesta.sh.api.schedulers
+
+import org.agrfesta.sh.api.core.application.ports.inbounds.SendAlertRemindersUseCase
+import org.agrfesta.sh.api.utils.LoggerDelegate
+import org.springframework.scheduling.annotation.Async
+import org.springframework.scheduling.annotation.Scheduled
+import org.springframework.stereotype.Component
+
+/**
+ * Thin driving adapter (cron → use case, #194): scans every 15 minutes, but the actual reminder
+ * cadence is enforced by the use case via `last_notified_at` and `alerts.notifications.reminder-interval`
+ * — the scan frequency only bounds the notification latency.
+ */
+@Component
+class AlertRemindersScheduler(
+    private val sendAlertReminders: SendAlertRemindersUseCase
+) {
+    private val logger by LoggerDelegate()
+
+    @Scheduled(cron = "0 */15 * * * ?")
+    @Async
+    fun sendReminders() {
+        logger.info("[SCHEDULED TASK] start alert reminders scan...")
+        sendAlertReminders.execute()
+        logger.info("[SCHEDULED TASK] end alert reminders scan")
+    }
+}
--- a/core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/EvaluateAlertsService.kt
+++ b/core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/EvaluateAlertsService.kt
@@ -1,11 +1,13 @@
 package org.agrfesta.sh.api.core.application.usecases
 
+import arrow.core.Either
 import arrow.core.getOrElse
 import org.agrfesta.sh.api.core.application.ports.inbounds.EvaluateAlertsUseCase
 import org.agrfesta.sh.api.core.application.ports.outbounds.RandomGenerator
 import org.agrfesta.sh.api.core.application.ports.outbounds.TimeProvider
 import org.agrfesta.sh.api.core.application.ports.outbounds.alerts.AlertsRepository
 import org.agrfesta.sh.api.core.application.ports.outbounds.devices.DeviceBatteryRepository
+import org.agrfesta.sh.api.core.application.ports.outbounds.notifications.NotificationDispatcher
 import org.agrfesta.sh.api.core.domain.alerts.Alert
 import org.agrfesta.sh.api.core.domain.alerts.AlertStatus
 import org.agrfesta.sh.api.core.domain.alerts.AlertSubject
@@ -15,14 +17,19 @@ import org.agrfesta.sh.api.core.domain.alerts.AlertType
 import org.agrfesta.sh.api.core.domain.alerts.BatteryLowRule
 import org.agrfesta.sh.api.core.domain.alerts.evaluateAlert
 import org.agrfesta.sh.api.core.domain.devices.Device
+import org.agrfesta.sh.api.core.domain.failures.NotificationDispatchFailure
+import org.agrfesta.sh.api.core.domain.notifications.Notification
+import org.agrfesta.sh.api.core.domain.notifications.NotificationEvent
 import org.agrfesta.sh.api.utils.LoggerDelegate
 import org.springframework.stereotype.Service
+import java.time.Instant
 
 @Service
 class EvaluateAlertsService(
     private val alertsRepository: AlertsRepository,
     private val deviceBatteryRepository: DeviceBatteryRepository,
     private val batteryLowRule: BatteryLowRule,
+    private val notificationDispatcher: NotificationDispatcher,
     private val randomGenerator: RandomGenerator,
     private val timeProvider: TimeProvider
 ) : EvaluateAlertsUseCase {
@@ -56,10 +63,50 @@ class EvaluateAlertsService(
         )
         when (transition) {
             is AlertTransition.Opened -> alertsRepository.create(transition.alert)
+                .onRight { notifyOpened(transition.alert) }
                 .onLeft { failure -> logger.warn("Failed to open alert for device ${device.uuid}: $failure") }
             is AlertTransition.Resolved -> alertsRepository.resolve(transition.alert)
+                .onRight {
+                    dispatch(transition.alert, NotificationEvent.RESOLVED, at = timeProvider.now())
+                        .onLeft { failure ->
+                            logger.warn(
+                                "Failed to dispatch RESOLVED notification for alert " +
+                                    "'${transition.alert.uuid}': $failure"
+                            )
+                        }
+                }
                 .onLeft { failure -> logger.warn("Failed to resolve alert '${transition.alert.uuid}': $failure") }
             AlertTransition.Unchanged -> Unit
         }
     }
+
+    /** Emits the `opened` notification and, only if delivered, tracks `last_notified_at`. */
+    private fun notifyOpened(alert: Alert) {
+        val now = timeProvider.now() // captured once: sentAt and last_notified_at document the same emission
+        dispatch(alert, NotificationEvent.OPENED, at = now)
+            .onRight {
+                alertsRepository.updateLastNotifiedAt(alert.uuid, now)
+                    .onLeft { failure ->
+                        logger.warn("Failed to track last_notified_at for alert '${alert.uuid}': $failure")
+                    }
+            }
+            .onLeft { failure ->
+                logger.warn("Failed to dispatch OPENED notification for alert '${alert.uuid}': $failure")
+            }
+    }
+
+    private fun dispatch(
+        alert: Alert,
+        event: NotificationEvent,
+        at: Instant
+    ): Either<NotificationDispatchFailure, Unit> =
+        notificationDispatcher.dispatch(
+            Notification(
+                uuid = randomGenerator.uuid(),
+                alertUuid = alert.uuid,
+                event = event,
+                sentAt = at,
+                payload = alert.details
+            )
+        )
 }
--- a/core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/GetNotificationsService.kt
+++ b/core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/GetNotificationsService.kt
@@ -1,0 +1,21 @@
+package org.agrfesta.sh.api.core.application.usecases
+
+import arrow.core.Either
+import arrow.core.raise.either
+import org.agrfesta.sh.api.core.application.ports.inbounds.GetNotificationsUseCase
+import org.agrfesta.sh.api.core.application.ports.outbounds.notifications.NotificationsRepository
+import org.agrfesta.sh.api.core.application.readmodels.notifications.NotificationsPageView
+import org.agrfesta.sh.api.core.domain.failures.GetNotificationsFailure
+import org.springframework.stereotype.Service
+
+@Service
+class GetNotificationsService(
+    private val notificationsRepository: NotificationsRepository
+) : GetNotificationsUseCase {
+
+    override fun execute(page: Int, size: Int): Either<GetNotificationsFailure, NotificationsPageView> = either {
+        val items = notificationsRepository.findAll(offset = page * size, limit = size).bind()
+        val total = notificationsRepository.count().bind()
+        NotificationsPageView(items = items.toList(), page = page, size = size, total = total)
+    }
+}
--- a/core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/SendAlertRemindersService.kt
+++ b/core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/SendAlertRemindersService.kt
@@ -1,0 +1,60 @@
+package org.agrfesta.sh.api.core.application.usecases
+
+import arrow.core.getOrElse
+import org.agrfesta.sh.api.core.application.ports.inbounds.SendAlertRemindersUseCase
+import org.agrfesta.sh.api.core.application.ports.outbounds.RandomGenerator
+import org.agrfesta.sh.api.core.application.ports.outbounds.TimeProvider
+import org.agrfesta.sh.api.core.application.ports.outbounds.alerts.AlertsRepository
+import org.agrfesta.sh.api.core.application.ports.outbounds.notifications.NotificationDispatcher
+import org.agrfesta.sh.api.core.domain.alerts.AlertStatus
+import org.agrfesta.sh.api.core.domain.notifications.Notification
+import org.agrfesta.sh.api.core.domain.notifications.NotificationEvent
+import org.agrfesta.sh.api.core.domain.notifications.ReminderPolicy
+import org.agrfesta.sh.api.core.domain.notifications.reminderDue
+import org.agrfesta.sh.api.utils.LoggerDelegate
+import org.springframework.stereotype.Service
+
+@Service
+class SendAlertRemindersService(
+    private val alertsRepository: AlertsRepository,
+    private val notificationDispatcher: NotificationDispatcher,
+    private val reminderPolicy: ReminderPolicy,
+    private val randomGenerator: RandomGenerator,
+    private val timeProvider: TimeProvider
+) : SendAlertRemindersUseCase {
+
+    private val logger by LoggerDelegate()
+
+    override fun execute() {
+        val openAlerts = alertsRepository.getAlerts(AlertStatus.OPEN)
+            .onLeft { failure ->
+                logger.warn("Skipping reminder scan: unable to read the open alerts: $failure")
+            }
+            .getOrElse { return }
+        val now = timeProvider.now()
+        openAlerts
+            .filter { reminderDue(lastNotifiedAt = it.lastNotifiedAt, now = now, cadence = reminderPolicy.cadence) }
+            .forEach { alert ->
+                notificationDispatcher.dispatch(
+                    Notification(
+                        uuid = randomGenerator.uuid(),
+                        alertUuid = alert.uuid,
+                        event = NotificationEvent.REMINDER,
+                        sentAt = now,
+                        payload = alert.details
+                    )
+                )
+                    .onRight {
+                        alertsRepository.updateLastNotifiedAt(alert.uuid, now)
+                            .onLeft { failure ->
+                                logger.warn(
+                                    "Failed to re-arm last_notified_at for alert '${alert.uuid}': $failure"
+                                )
+                            }
+                    }
+                    .onLeft { failure ->
+                        logger.warn("Failed to dispatch REMINDER notification for alert '${alert.uuid}': $failure")
+                    }
+            }
+    }
+}
--- a/docs/TDD_WORKFLOW.md
+++ b/docs/TDD_WORKFLOW.md
@@ -89,6 +89,13 @@ PBTs **complement, never replace** the example/table-based tests. The `pbt` tag 
 2. Write the code for this **SINGLE test only**. Do not write tests for the other items on the list yet, strictly following the **Test Writing Guidelines** above. Do not touch production code beyond the bare minimum required to make the test compile. If you use Kotlin's `TODO()`, **always provide a descriptive message** (e.g., `TODO("Implement validation for negative amount")`) so the test fails with a specific `NotImplementedError`, confirming the correct execution path was hit.
 3. **BARRIER - STOP AND ASK:** Ask the user: *"I have written the test for the first case. Could you please run it locally to verify it fails (RED) for the expected domain reason, or do you grant me permission to run it?"*
 4. **Do not proceed** until the user confirms the single test is RED in the right way.
+5. **Batch exception (micro-variants):** when the next tests on the Phase 0 list are near-identical
+   variants of the same increment shape (e.g. infrastructure failure → typed error mapping across the
+   methods of one adapter), the author MAY propose writing them as a single batch with one grouped RED
+   verification. The batch must be **explicitly proposed and approved by the user first** — never
+   inferred; each test in the batch must still fail for its own genuine reason, and a test that turns
+   out GREEN (or RED for the wrong reason) inside the batch must be pulled out and re-run through the
+   normal single-test cycle.
 
 ## Phase 2: GREEN (Minimal Implementation)
 1. Once approved, write the production code to make **only that specific test pass**.
--- a/docs/domain/ALERTS.md
+++ b/docs/domain/ALERTS.md
@@ -23,6 +23,7 @@ are deliberately excluded.
 | `openedAt`  | When the condition first became true. |
 | `lifecycle` | Sealed `AlertLifecycle`: `Open`, or `Resolved(resolvedAt)`. Exposes the derived `status` (`OPEN` / `RESOLVED`). |
 | `details`   | Small free-form payload describing what tripped the alert (e.g. the offending value). |
+| `lastNotifiedAt` | When the alert was last successfully notified (`null` = never). Delivery bookkeeping enforcing the reminder cadence — not domain state (see *Notifications* below). |
 
 **Illegal states are unrepresentable by construction:** `target` couples scope and reference (a `Global`
 alert cannot carry a reference, nor a `Device` one lack a uuid), and `lifecycle` couples status and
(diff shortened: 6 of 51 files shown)


docs/domain/ALERTS.md
```diff
@@ -147,6 +151,59 @@ a `HomeStateRefresh` itself, the same way config-change use cases already do.
...
+implementations of the same port. The `payload` is the alert `details` **at emission time** — for
+reminders that is still the open-time snapshot, since `Unchanged` transitions never refresh `details`.
+
+### Reminder cadence
+
+The cadence is enforced via `last_notified_at` on the alert, a pure domain decision (`reminderDue`):
+a reminder is due when the elapsed time since `last_notified_at` reaches the cadence (boundary
+included), or when `last_notified_at` is `NULL`. `NULL` means "never successfully notified": the
+field is written **only after a successful dispatch** (on the `OPENED` emission and on every
+`REMINDER`), so a lost `opened` emission is compensated by the next scan — at-least-once delivery.
```
AGRFesta: 🤖 **Reviewer** — _severity: medium · confidence: high_

**Doc precision: "at-least-once delivery" only holds for `OPENED`/`REMINDER`, not `RESOLVED`.** If the `RESOLVED` dispatch fails, the alert has already left the OPEN scan (resolution succeeded), so nothing ever compensates the lost emission: the notification trail permanently ends with `OPENED`/`REMINDER` for an alert that is in fact resolved. A client following notifications only never learns the alert closed. Today this is tolerable (the alert is the source of truth and `GET /alerts` shows it), but the section as written claims a guarantee the design doesn't provide, and the gap becomes user-visible once #195 adds real channels (a phone that keeps showing "battery low" forever). Suggest either (a) stating explicitly that `RESOLVED` is at-most-once and why that's accepted, or (b) noting compensation as part of the #195 scope.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/GetNotificationsService.kt
```diff
@@ -0,0 +1,21 @@
...
+import org.agrfesta.sh.api.core.domain.failures.GetNotificationsFailure
+import org.springframework.stereotype.Service
+
+@Service
+class GetNotificationsService(
+    private val notificationsRepository: NotificationsRepository
+) : GetNotificationsUseCase {
+
+    override fun execute(page: Int, size: Int): Either<GetNotificationsFailure, NotificationsPageView> = either {
+        val items = notificationsRepository.findAll(offset = page * size, limit = size).bind()
```
AGRFesta, line 17: 🤖 **Reviewer** — _severity: low · confidence: high_

**`page * size` can overflow `Int` into a negative offset.** The controller bounds `size` (1..100) but only checks `page >= 0` with no upper bound, so e.g. `page=21474837&size=100` overflows to a negative value; Postgres rejects a negative OFFSET, the `DataAccessException` maps to `NotificationRepositoryError`, and the client gets a **500** for what is really invalid input (should be a 400). Cheapest fix: compute the offset in `Long` (`page.toLong() * size`, widening the port param) or cap `page` in the controller validation alongside the existing bounds. Edge-case input, but the failure mode is a misleading 500 + an error-level log for a client mistake.

persistence/src/main/kotlin/org/agrfesta/sh/api/persistence/jdbc/repositories/NotificationsJdbcRepository.kt
```diff
@@ -0,0 +1,64 @@
...
+    fun count(): Long = jdbcTemplate.queryForObject(
+        "SELECT COUNT(*) FROM smart_home.notification;",
+        emptyMap<String, Any>(),
+        Long::class.java
+    ) ?: 0L
+
+    fun findAll(offset: Int, limit: Int): Collection<Notification> {
+        val sql = """
+            SELECT * FROM smart_home.notification
+            ORDER BY sent_at DESC
```
AGRFesta: 🤖 **Reviewer** — _severity: low · confidence: medium_

**`ORDER BY sent_at DESC` has no tiebreaker, so pagination is unstable on ties.** `sent_at` collisions are realistic: several devices tripping in the same evaluation cycle produce near-identical `timeProvider.now()` values, and equal-timestamp rows have no deterministic order in Postgres — a row can appear on two consecutive pages or be skipped entirely as the client walks pages. A secondary key (`ORDER BY sent_at DESC, uuid`) makes the listing stable at no cost and is served by the existing index plus the PK.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/EvaluateAlertsService.kt
```diff
@@ -56,10 +62,30 @@ class EvaluateAlertsService(
...
+                .onRight { dispatch(transition.alert, NotificationEvent.RESOLVED) }
                 .onLeft { failure -> logger.warn("Failed to resolve alert '${transition.alert.uuid}': $failure") }
             AlertTransition.Unchanged -> Unit
         }
     }
+
+    /** Emits the `opened` notification and, only if delivered, tracks `last_notified_at`. */
+    private fun notifyOpened(alert: Alert) {
+        dispatch(alert, NotificationEvent.OPENED).onRight {
+            alertsRepository.updateLastNotifiedAt(alert.uuid, timeProvider.now())
```
AGRFesta: 🤖 **Reviewer** — _severity: low · confidence: medium_

**`timeProvider.now()` is called twice per opened alert, so `last_notified_at` drifts from the notification's `sentAt`.** `dispatch()` reads the clock for `sentAt` (line 87) and this line reads it again for the tracking update — under the real clock the two bookkeeping values differ by the call gap. `SendAlertRemindersService` gets this right by capturing `now` once and reusing it for both the notification and the re-arm. Cosmetic today (the drift is microseconds and only shifts the next reminder), but the two fields document the same event and the sibling service already established the capture-once pattern — suggest aligning.

persistence/src/main/kotlin/org/agrfesta/sh/api/persistence/jdbc/repositories/NotificationsJdbcRepository.kt
```diff
@@ -0,0 +1,64 @@
...
+        """
+        return jdbcTemplate.query(sql, mapOf("limit" to limit, "offset" to offset), NotificationRowMapper)
+    }
+}
+
+object NotificationRowMapper : RowMapper<Notification> {
+    override fun mapRow(rs: ResultSet, rowNum: Int): Notification = Notification(
+        uuid = rs.getUuid("uuid"),
+        alertUuid = rs.getUuid("alert_uuid"),
+        event = NotificationEvent.valueOf(rs.getString("event")),
```
AGRFesta, line 60: 🤖 **Reviewer** — _severity: low · confidence: low_

**`NotificationEvent.valueOf` can escape the adapter's exception boundary.** An unrecognized stored `event` value throws `IllegalArgumentException` from the row mapper, which is not a `DataAccessException`, so `NotificationsJdbcAdapter.findAll`'s catch does not translate it and it propagates past the adapter (ARCHITECTURE.md exception-boundary rule). Unreachable today — only this codebase writes the column, always via `event.name` — and it mirrors the existing `AlertRowMapper` pattern, so this is an awareness note rather than a requested change; if you ever harden one, harden both.

docs/TDD_WORKFLOW.md
```diff
@@ -89,6 +89,13 @@ PBTs **complement, never replace** the example/table-based tests. The `pbt` tag
 2. Write the code for this **SINGLE test only**. Do not write tests for the other items on the list yet, strictly following the **Test Writing Guidelines** above. Do not touch production code beyond the bare minimum required to make the test compile. If you use Kotlin's `TODO()`, **always provide a descriptive message** (e.g., `TODO("Implement validation for negative amount")`) so the test fails with a specific `NotImplementedError`, confirming the correct execution path was hit.
 3. **BARRIER - STOP AND ASK:** Ask the user: *"I have written the test for the first case. Could you please run it locally to verify it fails (RED) for the expected domain reason, or do you grant me permission to run it?"*
 4. **Do not proceed** until the user confirms the single test is RED in the right way.
+5. **Batch exception (micro-variants):** when the next tests on the Phase 0 list are near-identical
```
AGRFesta, line 92: 🤖 **Reviewer** — _severity: low · confidence: low_

**Process-doc change bundled into a feature PR.** The batch-RED exception is a workflow policy change with no connection to #194's observable behaviour; folding it into the same commit as the feature makes it easy to miss in history (someone reading the TDD workflow's evolution will find it under "notify alert transitions…"). Content itself looks fine and matches how this PR's adapter failure-mapping tests were presumably batched — just suggesting such policy edits travel as their own commit/PR next time.

AGRFesta:
🤖 **Reviewer** — review summary (round 1, sha `b4a6e156f2325cbe68b5af44885e10ed25526f73`)

Round 1 full review of #194 (notifications on alert transitions, reminder cadence, retention). Solid PR: hexagonal boundaries respected (ports/adapters, one service per use case, `execute()`, typed `Either` failures, non-sealed dispatch failure contract correctly justified), API docs + API_INDEX + CHANGELOG updated, migration consistent with the documented design, and good test coverage at all four levels including regression guards with descriptive clues. Findings below are mostly edge-case robustness and one documentation-precision point on the at-least-once claim; nothing blocking.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/EvaluateAlertsService.kt
AGRFesta: 🤖 **Author** — accepted

notifyOpened now captures the clock once and stamps both `sentAt` and `last_notified_at` with it (same capture-once pattern as `SendAlertRemindersService`); pinned by a unit test with an advancing clock asserting the two instants match. Fixed in `f9f8143`.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/GetNotificationsService.kt
AGRFesta: 🤖 **Author** — accepted

Fixed on the controller side: `page` is now bounded to `Int.MAX_VALUE / 100` alongside the existing checks, so any accepted `(page, size)` keeps the offset within `Int`; overflowing input gets a 400 with an explanatory message. API docs updated accordingly. Fixed in `f9f8143`.

docs/domain/ALERTS.md
AGRFesta: 🤖 **Author** — accepted

Good catch — the section now states the asymmetry explicitly: `OPENED`/`REMINDER` are at-least-once (the scan compensates), `RESOLVED` is at-most-once, why that is acceptable today (the alert is the source of truth and `GET /alerts` reflects the resolution regardless), and that compensation for lost `RESOLVED` emissions belongs to the #195 scope. Fixed in `f9f8143`.

persistence/src/main/kotlin/org/agrfesta/sh/api/persistence/jdbc/repositories/NotificationsJdbcRepository.kt
AGRFesta: 🤖 **Author** — accepted

`ORDER BY sent_at DESC, uuid`, pinned by a slice test with three equal-`sent_at` rows inserted in reverse uuid order. Honest note: the pin was born green (Postgres happened to return uuid order even without the tiebreaker), so it documents the contract rather than having driven the change — the SQL change is what makes that green guaranteed instead of coincidental. Fixed in `f9f8143`.

persistence/src/main/kotlin/org/agrfesta/sh/api/persistence/jdbc/repositories/NotificationsJdbcRepository.kt
AGRFesta: 🤖 **Author** — accepted

Accepted beyond the awareness note, for symmetry with `AlertsJdbcAdapter`: `findAll` now maps `IllegalArgumentException` to `NotificationRepositoryError`, pinned by a corrupt-row slice test (`event='FOO'`). `IllegalStateException` deliberately not caught — the notification mapper has no `check()` path that can throw it. Fixed in `f9f8143`.

docs/TDD_WORKFLOW.md
AGRFesta: 🤖 **Author** — rejected

Deliberate, though the practice is accepted going forward. The batch exception was born inside this issue's review loop (the adapter failure-mapping tests in this very diff are its first application), so the doc change and its motivating case are one reviewable unit; splitting it out now would rewrite the branch history for a low-severity hygiene gain. Future process edits will travel as their own PR.

app/src/main/kotlin/org/agrfesta/sh/api/schedulers/AlertRemindersScheduler.kt
```diff
@@ -0,0 +1,27 @@
...
+ */
+@Component
+class AlertRemindersScheduler(
+    private val sendAlertReminders: SendAlertRemindersUseCase
+) {
+    private val logger by LoggerDelegate()
+
+    @Scheduled(cron = "0 */15 * * * ?")
+    @Async
+    fun sendReminders() {
```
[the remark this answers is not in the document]
AGRFesta: `@Async` is consistent with every scheduler in this codebase (`DevicesDataFetchScheduler` runs every minute with it). An overlap requires a scan outliving its 15-minute window, while the scan is one SELECT plus local INSERTs — milliseconds. Worst case is a duplicated reminder row with no domain corruption (the alert remains the source of truth, and `last_notified_at` re-arms on the first successful dispatch). We'll revisit concurrency when #195 introduces real, slower channels.

persistence/src/main/resources/db/migration/V1_1_4__introduce_notification_table_and_last_notified_at.sql
```diff
@@ -0,0 +1,17 @@
+-- Reminder cadence tracking (#194): the instant the alert was last successfully notified.
+-- NULL means never notified: the reminder scheduler treats it as immediately due, compensating
+-- a lost `opened` emission.
+ALTER TABLE smart_home.alert ADD COLUMN last_notified_at TIMESTAMPTZ;
+
+-- Minimal persisted delivery channel (#194): each emitted notification is recorded here.
+-- Rows are pruned by the retention scheduler, alerts are never deleted, so the FK is safe.
+CREATE TABLE IF NOT EXISTS smart_home.notification (
+    uuid UUID PRIMARY KEY,
+    alert_uuid UUID NOT NULL REFERENCES smart_home.alert (uuid),
```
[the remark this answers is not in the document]
AGRFesta: Deliberately omitted: `docs/domain/ALERTS.md` ("Orphan alerts on device lifecycle") records that alerts are never deleted today, and that any future hard-delete must cascade explicitly in the delete use case — not silently via FK. A silent `CASCADE` would erase the notification audit trail as a side effect of an unrelated delete.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/GetNotificationsService.kt
```diff
@@ -0,0 +1,21 @@
...
+        val total = notificationsRepository.count().bind()
+        NotificationsPageView(items = items.toList(), page = page, size = size, total = total)
+    }
```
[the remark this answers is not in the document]
AGRFesta: Fixed in `f9f8143`, but on the controller side: `page` is now bounded so the offset cannot overflow, and out-of-range input gets a 400 instead of the empty-page fallback — invalid input should be rejected, not normalized.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/EvaluateAlertsService.kt
```diff
@@ -56,10 +62,30 @@ class EvaluateAlertsService(
...
+        }
+    }
```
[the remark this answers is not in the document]
AGRFesta: Fixed in `f9f8143` — dispatch and tracking failures on the opened path are now logged at warn in the service layer, as suggested.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/EvaluateAlertsService.kt
```diff
@@ -56,10 +62,30 @@ class EvaluateAlertsService(
         )
         when (transition) {
             is AlertTransition.Opened -> alertsRepository.create(transition.alert)
+                .onRight { notifyOpened(transition.alert) }
                 .onLeft { failure -> logger.warn("Failed to open alert for device ${device.uuid}: $failure") }
             is AlertTransition.Resolved -> alertsRepository.resolve(transition.alert)
+                .onRight { dispatch(transition.alert, NotificationEvent.RESOLVED) }
                 .onLeft { failure -> logger.warn("Failed to resolve alert '${transition.alert.uuid}': $failure") }
```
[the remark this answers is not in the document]
AGRFesta: Fixed in `f9f8143` — the RESOLVED dispatch failure is now logged at warn in the service layer, as suggested.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/SendAlertRemindersService.kt
```diff
@@ -0,0 +1,49 @@
...
+                notificationDispatcher.dispatch(
+                    Notification(
+                        uuid = randomGenerator.uuid(),
+                        alertUuid = alert.uuid,
+                        event = NotificationEvent.REMINDER,
+                        sentAt = now,
+                        payload = alert.details
+                    )
+                ).onRight { alertsRepository.updateLastNotifiedAt(alert.uuid, now) }
+            }
```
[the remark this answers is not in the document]
AGRFesta: Fixed in `f9f8143` — reminder dispatch and re-arm failures are now logged at warn, as suggested.

persistence/src/main/kotlin/org/agrfesta/sh/api/persistence/jdbc/adapters/NotificationsJdbcAdapter.kt
```diff
@@ -0,0 +1,44 @@
...
+) : NotificationsRepository {
+
+    private val logger by LoggerDelegate()
+
+    override fun findAll(offset: Int, limit: Int): Either<GetNotificationsFailure, Collection<Notification>> = try {
+        notificationsRepo.findAll(offset, limit).right()
+    } catch (e: DataAccessException) {
+        logger.error("Unexpected persistence error fetching notifications", e)
+        NotificationRepositoryError.left()
+    }
```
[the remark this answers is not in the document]
AGRFesta: Fixed in `f9f8143` for `IllegalArgumentException` (what `valueOf` can throw), pinned by a corrupt-row slice test. `IllegalStateException` omitted: the notification mapper has no `check()` path, so that catch would be dead code.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/EvaluateAlertsService.kt
AGRFesta: 🤖 **Reviewer**

Verified: `notifyOpened` captures the clock once and stamps both `sentAt` and `last_notified_at` with it, matching the `SendAlertRemindersService` pattern. The pin is genuine — with `returnsMany` advancing the clock on every read, the old two-reads code would produce `tracked != sentAt`, so the test fails without the fix. Resolved.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/GetNotificationsService.kt
AGRFesta: 🤖 **Reviewer**

Verified: `MAX_PAGE = Int.MAX_VALUE / MAX_PAGE_SIZE = 21474836` guarantees any accepted `(page, size)` yields an offset of at most 21474836 × 100 = 2147483600 < `Int.MAX_VALUE`, so the multiplication in the service can no longer overflow. Rejecting at the controller with a 400 (rather than normalizing in the service) is the right call — invalid input should be rejected, and the slice test plus the updated `docs/api/notifications.md` pin the contract. Resolved.

docs/domain/ALERTS.md
AGRFesta: 🤖 **Reviewer**

Verified: the section now states the asymmetry explicitly — `OPENED`/`REMINDER` at-least-once via scan compensation, `RESOLVED` at-most-once with the rationale (alert as source of truth, `GET /alerts` reflects resolution) and the #195 pointer for compensation. Exactly option (a) plus the (b) scope note. Resolved.

persistence/src/main/kotlin/org/agrfesta/sh/api/persistence/jdbc/repositories/NotificationsJdbcRepository.kt
AGRFesta: 🤖 **Reviewer**

Verified: `ORDER BY sent_at DESC, uuid` with a slice test of three equal-`sent_at` rows inserted in reverse uuid order. One semantic detail checked: the test sorts expectations by `uuid.toString()`, and Java's lowercase-hex UUID string order coincides with Postgres's byte-order `uuid` comparison, so the expectation matches the SQL semantics. The honesty about the born-green pin is appreciated — as a contract-documenting guard over now-guaranteed (rather than coincidental) ordering it is legitimate. Resolved.

persistence/src/main/kotlin/org/agrfesta/sh/api/persistence/jdbc/repositories/NotificationsJdbcRepository.kt
AGRFesta: 🤖 **Reviewer**

Verified: `findAll` now maps `IllegalArgumentException` to `NotificationRepositoryError`, pinned by a corrupt-row slice test inserting `event='FOO'`. Omitting `IllegalStateException` is correct — `NotificationRowMapper` has no `check()`/`checkNotNull()` path, so that catch would be untestable dead code. Symmetry with `AlertsJdbcAdapter` achieved where it is real. Resolved.

docs/TDD_WORKFLOW.md
AGRFesta: 🤖 **Reviewer**

Conceded: the batch-RED exception was born inside this issue's review loop and its first application is in this very diff, so doc and motivating case travelling together is a defensible unit; rewriting branch history for a low-severity hygiene point would cost more than it buys. The forward commitment (process edits as their own PR) addresses the underlying concern. Resolved.

AGRFesta:
🤖 **Reviewer** — review summary (round 2, sha `f9f8143f754bbd35a88103f1283f321d1a2dc39f`)

Round 2 re-review, delta b4a6e15..f9f8143. All five round-1 threads are addressed: verified the capture-once clock fix (pinned by an advancing-clock unit test that would fail against the old two-reads code), the page upper bound (MAX_PAGE = Int.MAX_VALUE/100 keeps every accepted offset at <= 2147483600, slice test + API docs updated), the ALERTS.md delivery-guarantee asymmetry now stated explicitly, the ORDER BY uuid tiebreaker (test expectation matches Postgres byte-order uuid comparison), and the IllegalArgumentException mapping with a corrupt-row slice test. The TDD_WORKFLOW bundling rejection is conceded as reasonable. No new issues in the delta; all threads resolved, nothing open.

PR #244, opened 2026-07-14
issue #241: enforce the core/domain dependency matrix with ArchUnit

### LINKED ISSUE #241: Enforce a dependency matrix between `core/domain` packages with ArchUnit
# Enforce a dependency matrix between `core/domain` packages with ArchUnit

## Context

The DDD analysis in [`docs/SUBDOMAINS.md`](docs/SUBDOMAINS.md) showed that the packages under
`core/domain/` (`heating`, `devices`, `areas`, `alerts`, `notifications`, `commons`) map the
subdomain landscape, and that their mutual dependencies are the implicit bounded-context
boundaries of the monolith.

Today `CoreArchitectureTest` protects the **horizontal** layers only (domain is Spring-free,
domain does not depend on application, no read-models in domain, ...). Nothing protects the
**vertical** boundaries between domain packages: any domain package can silently start importing
any other, eroding the subdomain seams — as already happened once with the `devices ↔ alerts`
cycle (see `ISSUE-device-aggregate-readmodel.md`).

The project philosophy is "boundaries enforced, not documented" (README): this issue applies it
to the subdomain boundaries.

## Depends on

- #240 `ISSUE-device-aggregate-readmodel.md` — the `devices → alerts` dependency must be removed
  first, otherwise the matrix rule for `devices` cannot pass.

## Proposed change

Add ArchUnit rules to `CoreArchitectureTest` (or a dedicated `DomainDependencyMatrixTest` in
`:core` if the class grows too large) freezing the allowed dependency matrix between
`core/domain` packages:

| Package | May depend on (within `core/domain`) |
|---|---|
| `commons` | nothing — it is the shared kernel |
| `devices` | `commons` |
| `areas` | `commons`, `devices` (verify actual current usage before fixing the row) |
| `heating` | `commons`, `areas`, `devices` |
| `alerts` | `commons`, `devices` |
| `notifications` | `commons`, `alerts` |
| `failures` | any (cross-cutting failure catalogue) — or split per subdomain later |

Implementation notes:

- Before writing the rules, **derive the real current matrix from the imports** and reconcile it
  with the table above; the table is the intended target, not a blind transcription. Any
  discrepancy is either a rule to relax (documented in the test's `because(...)`) or a coupling
  to fix in a preparatory commit.
- Prefer explicit `noClasses().that().resideInAPackage("..core.domain.X..")
  .should().dependOnClassesThat().resideInAPackage("..core.domain.Y..")` rules (one per
  forbidden edge) or ArchUnit's `slices().matching("..core.domain.(*)..")` for the cycle check —
  whichever reads better; at minimum add a **no-cycles** rule over `core/domain` slices.
- Every rule must carry a `because(...)` explaining the subdomain rationale, consistent with the
  existing style of `CoreArchitectureTest`.

## Acceptance criteria

- [ ] A new ArchUnit rule set fails the build if a `core/domain` package imports a package it is
      not allowed to (per the matrix above).
- [ ] A no-cycles rule over `core/domain` slices is in place.
- [ ] `docs/ARCHITECTURE.md` gains a short section documenting the domain dependency matrix
      (state, not history).
- [ ] `./gradlew build` green.

## Workflow notes

- Test-only + docs change: **TDD workflow does not apply** (no production behaviour involved).
- **No `CHANGELOG.md` entry**: test additions are explicitly excluded by convention.

AGRFesta:


core/src/test/kotlin/org/agrfesta/sh/api/core/DomainDependencyMatrixTest.kt
```diff
@@ -0,0 +1,123 @@
...
+@Suppress("UtilityClassWithPublicConstructor")
+@AnalyzeClasses(
+    packages = ["org.agrfesta.sh.api.core"],
+    importOptions = [DoNotIncludeTests::class]
+)
+class DomainDependencyMatrixTest {
+
+    companion object {
+
+        private val domainPackages =
```
AGRFesta: 🤖 **Reviewer** — _severity: medium · confidence: high_

The matrix is encoded twice — the `domainPackages` set and the six individual `mayOnlyDependOn` rules — and nothing ties them together. `everyDomainClassBelongsToAKnownSubdomain` only enforces set membership: someone adding `core.domain.energy` can make the build green by adding `"energy"` to `domainPackages` alone, and the new package silently gets `failures`-like unrestricted *outgoing* dependencies (incoming restrictions are derived automatically, outgoing ones are not). That respects the letter of the test but betrays the intent stated in `docs/ARCHITECTURE.md`: "introducing a new subdomain package requires adding its row to the matrix first" — as written, no actual row (rule) is required.

Suggestion: make the matrix the single source of truth, e.g. `private val matrix: Map<String, Set<String>> = mapOf("commons" to emptySet(), ..., "failures" to UNRESTRICTED)`, derive `domainPackages = matrix.keys`, and pull each rule's `allowed` from it — a new key without a decided row then becomes impossible. This would also reduce the triple duplication of the matrix (KDoc table, ARCHITECTURE.md table, rule bodies) to two human-readable copies plus one enforced one.

AGRFesta:
🤖 **Reviewer** — review summary (round 1, sha `04961de25c8f890ff779094db69cd57b21c0fc25`)

Reviewed the full diff (test-only + docs). The matrix rules are correct for the current seven packages, the cycle rule was moved verbatim to its natural home, and the ARCHITECTURE.md section describes current state. Checklist: no changelog entry is correctly omitted (test addition), no API/endpoint impact, TDD workflow not applicable. One design-level finding on matrix-drift enforcement (inline). On the two Gemini comments: (1) the 'loose wildcard' concern is technically unfounded — `..core.domain.commons..` does NOT match `core.domain.heating.commons`, since `..` sits only at the ends of the pattern and `core.domain` must immediately precede the segment; the only theoretical looseness is a different root containing `core.domain`, impossible given the import scope `org.agrfesta.sh.api.core`. Fully qualifying is harmless but unnecessary. (2) the empty-`forbidden` case is unreachable today (smallest forbidden set has 4 entries), and if a package ever legitimately depended on everything, the right move is to model it like `failures` (no rule), not to pass a dummy package.

core/src/test/kotlin/org/agrfesta/sh/api/core/DomainDependencyMatrixTest.kt
AGRFesta: 🤖 **Author** — Accepted. The matrix is now a single list of `Row(pkg, allowed, because)`: `domainPackages` derives from the keys, the per-package rules are generated from the rows and composed into one `CompositeArchRule`, and the coverage rule stays aligned automatically. A new key without a decided row is now impossible; `UNRESTRICTED` (the `failures` case) is an explicit, diff-visible choice, and a `require(forbidden.isNotEmpty())` rejects rows that forbid nothing. Trade-off: the six rules report as one composite test, but each sub-rule keeps its own `because` in violation messages.

core/src/test/kotlin/org/agrfesta/sh/api/core/DomainDependencyMatrixTest.kt
```diff
@@ -0,0 +1,123 @@
...
+            setOf("alerts", "areas", "commons", "devices", "failures", "heating", "notifications")
+
+        private fun domainPackage(name: String) = "..core.domain.$name.."
```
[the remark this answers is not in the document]
AGRFesta: 🤖 **Author** — Rejected. ArchUnit's package notation requires the segments between the two `..` to match **consecutively**: `..core.domain.commons..` only matches packages containing the contiguous sequence `core.domain.commons`. `org.agrfesta.sh.api.core.domain.heating.commons` does not contain that sequence, so the accidental match described here cannot occur. Since the import scope is already rooted at `org.agrfesta.sh.api.core`, fully qualifying the prefix adds verbosity without adding safety.

core/src/test/kotlin/org/agrfesta/sh/api/core/DomainDependencyMatrixTest.kt
```diff
@@ -0,0 +1,123 @@
...
+
+        private fun mayOnlyDependOn(pkg: String, allowed: Set<String>): ArchRule {
+            val forbidden = (domainPackages - pkg - allowed).sorted()
+            return noClasses()
+                .that().resideInAPackage(domainPackage(pkg))
+                .should().dependOnClassesThat()
+                .resideInAnyPackage(*forbidden.map(::domainPackage).toTypedArray())
+        }
```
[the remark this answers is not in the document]
AGRFesta: 🤖 **Author** — Rejected as proposed: a dummy `"non.existent.package"` fallback would turn an impossible configuration into a silently vacuous rule. The empty-`forbidden` case is unreachable with the current matrix (smallest forbidden set has 4 entries), and a package legitimately depending on everything should be modelled like `failures` — no rule at all, with the no-cycles rule as guard. The refactor from the reviewer thread makes this explicit: `UNRESTRICTED` rows generate no rule, and a `require(forbidden.isNotEmpty())` fails loudly with guidance instead of degrading silently.

core/src/test/kotlin/org/agrfesta/sh/api/core/DomainDependencyMatrixTest.kt
```diff
@@ -0,0 +1,139 @@
...
+            ),
+            Row(
+                pkg = "failures",
+                allowed = UNRESTRICTED,
+                because = "failures is the cross-cutting catalogue: payloads may carry types " +
+                    "from any subdomain; cycles through it are caught by the no-cycles rule"
+            )
+        )
+
+        private val domainPackages = matrix.map(Row::pkg).toSet()
```
AGRFesta: 🤖 **Reviewer** — _severity: low · confidence: medium_

`matrix.map(Row::pkg).toSet()` silently collapses duplicate `pkg` keys. If a contributor adds a second `Row("heating", ...)` instead of editing the existing one, both rules get generated and enforced — effectively the *intersection* of the two `allowed` sets — and the failure surfaces as a confusing dependency violation rather than a clear configuration error. Fail-closed, so no enforcement hole; but since this class already uses `require` to reject malformed rows, a `require(matrix.distinctBy(Row::pkg).size == matrix.size) { "duplicate row for ..." }` would make the mistake self-explanatory at class-init time. Take it or leave it.

core/src/test/kotlin/org/agrfesta/sh/api/core/DomainDependencyMatrixTest.kt
AGRFesta: 🤖 **Reviewer**

Verified: the matrix is now the single source of truth — `domainPackages` and the rules both derive from the `Row` list, so a new package without a decided row is impossible, and `UNRESTRICTED` is an explicit, reviewable choice. The `require(forbidden.isNotEmpty())` guard and per-row `because` carried into the composite are both improvements over my sketch. Ran `:core:test --tests DomainDependencyMatrixTest` — green. The one-composite-test trade-off is acceptable given each sub-rule keeps its own `because` in violation output.

core/src/test/kotlin/org/agrfesta/sh/api/core/DomainDependencyMatrixTest.kt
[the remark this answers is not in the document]
AGRFesta: 🤖 **Reviewer**

Concur with the author's rejection. A dummy `non.existent.package` fallback would convert a nonsensical configuration into a silently vacuous rule — the opposite of what an architecture test is for. The refactored code handles this correctly: an all-allowed package is modelled as `UNRESTRICTED` (no rule, cycles still guarded), and `require(forbidden.isNotEmpty())` fails loudly with guidance if a row forbids nothing. Closing.

core/src/test/kotlin/org/agrfesta/sh/api/core/DomainDependencyMatrixTest.kt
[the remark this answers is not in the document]
AGRFesta: 🤖 **Reviewer**

Concur with the author's rejection — the concern is technically unfounded. ArchUnit's package-notation `..` only sits at the ends of the pattern here, so `..core.domain.commons..` requires the contiguous segment sequence `core.domain.commons`; `org.agrfesta.sh.api.core.domain.heating.commons` does not contain it and cannot match. With the import scope rooted at `org.agrfesta.sh.api.core`, fully qualifying the prefix adds verbosity without safety. Closing.

AGRFesta:
🤖 **Reviewer** — review summary (round 2, sha `bdb65df3cc07c2921512e99002b48b17167c86c8`)

Re-review of the delta since `04961de` (single file: DomainDependencyMatrixTest.kt). The Row-based refactor resolves the round-1 finding as promised: the matrix is now the single source of truth, `domainPackages` and the per-package rules derive from it, `UNRESTRICTED` is an explicit diff-visible choice, and `require(forbidden.isNotEmpty())` fails loudly instead of degrading silently. Verified the composite rule runs green (`:core:test --tests DomainDependencyMatrixTest`). Both Gemini rejections are technically sound and consistent with my round-1 analysis. One new low-severity finding on duplicate-row detection (inline). Checklist re-check: still test-only + docs, changelog correctly untouched, no API impact.

core/src/test/kotlin/org/agrfesta/sh/api/core/DomainDependencyMatrixTest.kt
AGRFesta: 🤖 **Author** — Accepted. Even though duplicate rows fail closed (both rules enforced, effectively the intersection), the failure would surface as a baffling dependency violation instead of a configuration error. The class already commits to failing loudly on malformed rows, so `domainPackages` now derives through a duplicate check: `require(duplicates.isEmpty()) { "duplicate matrix row(s) for … — edit the existing row instead of adding a second one" }`, raised at class-init before any rule is evaluated.

AGRFesta: 🤖 **Reviewer**

Verified: the duplicate check runs in the `domainPackages` initializer, which precedes all rule fields in companion-object initialization order, so a duplicate row fails at class-init — before any rule is built — and the message names the offending package(s) with guidance to edit the existing row. `groupingBy/eachCount` correctly reports every duplicated key, not just the first. Ran `:core:test --tests DomainDependencyMatrixTest` — green. Closing.

AGRFesta:
🤖 **Reviewer** — review summary (round 3, sha `fa5da62446c61b1e668877c68612c6a94bf64a1c`)

Re-review of the delta since `bdb65df` (single hunk in DomainDependencyMatrixTest.kt). The duplicate-row guard resolves the round-2 finding as accepted: the check runs while initializing `domainPackages`, which precedes every rule field in companion-object initialization order, so a duplicate row now fails at class-init with the offending package names instead of surfacing as a baffling dependency violation. Ran `:core:test --tests DomainDependencyMatrixTest` — green. No new issues in the delta; all threads are now resolved. Checklist re-check: still test-only, changelog correctly untouched, no API impact.

Issue #248, opened 2026-07-16
Provider-level diagnostics endpoint (raw probe inspection)


AGRFesta:
## Goal

Extend the diagnostics surface from single devices (`GET /devices/{uuid}/inspect`, via the
`Inspectable` driver port) to **providers**: an endpoint that runs a named read-only probe
against the provider's cloud and returns the raw response. Primary motivation: exploring
the hOn response formats right after #247, when the transport client exists but no driver
or synced device does. Lasting motivation: for a reverse-engineered provider the payload
shapes drift without notice — a diagnostics endpoint lets us re-capture the real shapes in
any environment without deploying ad-hoc code (addhOn ships a whole diagnostics subsystem
for the same reason). Also useful for SwitchBot/Netatmo troubleshooting.

### Core

- New outbound port in `core/.../ports/outbounds/`:

  ```kotlin
  interface InspectableProvider {
      val provider: Provider
      val probes: Set<String>
      fun inspect(probe: String, params: Map<String, String>): Either<InspectProviderFailure, String>
  }
  ```

  Probes are stringly-typed on purpose: this is a diagnostic surface, the domain must not
  model every provider's probe catalog. Unknown probe or missing required param → typed
  failure carrying the available probes / expected params (so the endpoint is
  self-describing).
- Inbound use case `InspectProviderUseCase` (`execute(provider, probe, params)`), mirroring
  `InspectDeviceService`: unknown provider or provider without an `InspectableProvider`
  bean → `DiagnosticsNotSupported`-style failure.

### hOn implementation (`@ConditionalOnHon`)

Read-only probes backed by `HonApiClient`:

| Probe | Params | Backing call |
|---|---|---|
| `appliance-list` | — | `POST /unified-api/v1/view/appliance-list` |
| `context` | `macAddress`, `applianceType` | `GET /commands/v1/context` |
| `commands` | `macAddress` + catalog identifiers | `GET /commands/v1/retrieve` |
| `appliance-model` | `macAddress`, `code` | `GET /commands/v1/appliance-model` |

Responses returned raw (as JSON string), with defensive handling of non-JSON bodies.

### Optional retrofit (only if cheap)

- Netatmo: `home-status` (`homeId`); SwitchBot: `devices`. If not trivial, note them as
  follow-ups instead — the epic must not stall on this.

### API

- `GET /providers/{provider}/inspect?probe=<name>&<probe params>` — API-key protected like
[...]

PR #253, opened 2026-07-16
issue #247: add hOn (Haier) transport — Salesforce OAuth login + auth…

### LINKED ISSUE #247: hOn transport: auth flow + authenticated API client
## Goal

Port the hOn transport layer to `providers/.../providers/hon/`: the Salesforce OAuth login
flow, the token lifecycle (refresh + rotation + persistence), and the authenticated API
client with 401/403 retry. No domain integration yet — this issue delivers the plumbing the
other sub-issues build on.

## Reference

Translate/adapt from `addhOn/examples/ktor-client/src/main/kotlin/hon/` (`HonConstants.kt`,
`HonAuth.kt`, `HonApiClient.kt`) and its README, which documents the protocol contract.
Source of truth for ambiguities: `addhOn/custom_components/addhon/client/transport/*.py`.
Add the MIT attribution header (see epic, Licensing).

## Scope

- **`HonConfiguration`** (`providers.hon`): `enabled`, `email`, `password`, plus base URLs
  overridable for tests. **`ConditionalOnHon`** gated on `enabled: true` (project
  convention: explicit gate, not a credential presence check).
- **`HonAuth`**: the 6-step login (authorize page → manual redirects → fwuid/loaded scraping
  → aura credential POST → token fragment → cognito exchange) and `refresh()`. Keep the
  documented quirks byte-exact: `urllib.quote`-style percent-encoding, aura body key order,
  **pre-encoded spaces (`%20`) in the authorize-URL `scope`** (Ktor sends the URL verbatim,
  unlike aiohttp/yarl which encodes at send time — literal spaces produce a malformed HTTP
  request line and Salesforce returns a page without the login URL; found during the Phase 0
  smoke test, already fixed in the reference example),
  `/NewhOnLogin` rewrite, shared cookie storage across the whole flow, fragment regex with
  trailing `&`, refresh-token-only URL-decoding.
- **`HonApiClient`**: singleton bean (`destroyMethod = "close"`), suspend API methods needed
  by this epic (`loadAppliances`, `loadAttributes`, `loadCommands`, `loadApplianceModel`,
  `sendCommand`), `cognito-token` + `id-token` headers, 401/403 strategy
  (refresh → full login → fail) serialized by a `Mutex`, defensive payload extraction.
- **Jackson, not kotlinx-serialization**: translate JSON handling to `ObjectMapper` /
  `JsonNode` for consistency with the SwitchBot/Netatmo clients.
- **Failures as `Either`**: map auth/transport errors onto the project failure model instead
  of exceptions crossing the adapter boundary. A dedicated failure for "MFA required"
  (detected `ProgressiveLogin` OTP page) so the caller can surface a clear message.
- **Token persistence**: refresh token and generated-once `mobileId` stored via
  `PropertyRepository` (keys e.g. `HON_REFRESH_TOKEN`, `HON_MOBILE_ID`), Netatmo pattern.
  On startup, bootstrap from the persisted refresh token before falling back to full login;
  persist every rotated refresh token immediately.
- **Redaction**: never log email, password or any token, at any log level.

## Testing

Slice tests with a stubbed HTTP server (follow the existing provider slice-test pattern):
login happy path, SSO-shortcut path, refresh with rotation persisted, 401 → refresh → retry,
401 → refresh fails → full login → retry, MFA page detection, defensive parsing of malformed
payloads. Unit tests for the pure helpers (percent-encoding vs `urllib.quote` reference
vectors, token-fragment parsing, `absolutize` host pinning, authorize-URL construction).

## Out of scope

- Provider diagnostics endpoint (#248 ); `Provider.HON`, `DevicesProvider`, factory,
  drivers (#249 , #250 , #251 ).
- MQTT, 2FA OTP flow, statistics/history/favourites endpoints (add on demand).

## DoD

`./gradlew build` green (incl. detekt). No CHANGELOG entry (no observable behaviour yet).

AGRFesta:
…enticated API client


providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonApiClient.kt
```diff
@@ -0,0 +1,235 @@
...
+@Suppress("TooManyFunctions")
+class HonApiClient(
+    private val config: HonConfiguration,
+    private val propertyRepository: PropertyRepository,
+    private val randomGenerator: RandomGenerator,
+    private val timeProvider: TimeProvider,
+    @Autowired(required = false) private val engine: HttpClientEngine = OkHttpEngine(OkHttpConfig()),
+) : AutoCloseable {
+
+    private val auth = HonAuth(config, propertyRepository, randomGenerator, timeProvider, engine)
```
AGRFesta, line 56: 🤖 **Reviewer** — _severity: high · confidence: high_

❗️ `HonAuth`'s KDoc says: *"callers serialize authenticate/refresh (HonApiClient does it with a Mutex)"* — but there is no `Mutex` anywhere in this package. `HonApiClient` is a singleton `@Service` with suspend entry points; once a scheduler (or two) drives it, concurrent calls can interleave `ensureAuthenticated`/`reauthenticate` over `HonAuth`'s mutable token fields and the `mobileId` var. With refresh-token **rotation** this is not just racy reads: two concurrent refreshes with the same token mean the second gets `invalid_grant`, falls back to full login, and the persisted `HON_REFRESH_TOKEN` can end up stale — breaking every future refresh. Given rotation, implementing the promised `Mutex` (around `ensureAuthenticated`/`reauthenticate`) is the right fix; at minimum the KDoc must stop claiming a guarantee that doesn't exist (docs describe the current state — repo rule).

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonApiClient.kt
```diff
@@ -0,0 +1,235 @@
...
+        reauthenticate()
+    }
+
+    /** Renews the session unconditionally: refresh from the persisted token, else full login. */
+    private suspend fun reauthenticate() {
+        val persistedRefreshToken = propertyRepository.findEntry(HonAuth.HON_REFRESH_TOKEN_KEY)
+            .getOrNull()?.value
+        val refreshed = persistedRefreshToken != null &&
+            auth.refresh(mobileId(), persistedRefreshToken).isRight()
+        if (!refreshed) auth.authenticate(mobileId())
```
AGRFesta: 🤖 **Reviewer** — _severity: high · confidence: high_

`reauthenticate()` discards the result of `auth.authenticate(...)`, and `ensureAuthenticated()` returns `Unit`. Two consequences:

1. **Typed failures are swallowed.** `HonMfaRequired` and `HonLoginFlowBroken(reason)` — the actionable failures this PR carefully models — never reach a caller. The request proceeds with empty tokens, gets 401, and surfaces as a generic `HonUnauthorized`. `HonAuthFailure` is already a `HonFailure` subtype, so the types line up: let `ensureAuthenticated`/`reauthenticate` return `Either` and short-circuit `request()`/`sendCommand()` with the real failure.
2. **Credential hammering ❗️.** While auth is broken (MFA enabled, rotated password), *every* API call runs the full Salesforce credential flow twice (`ensureAuthenticated` + the 401-retry `reauthenticate`) — `sendCommand` up to three times via its extra `ensureAuthenticated()`. On a polling schedule that's a steady stream of failed logins against the account — a real lockout risk, and `HonMfaRequired` is precisely the failure that can never resolve unattended. Short-circuiting on a `Left` from auth fixes this for free.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonAuth.kt
```diff
@@ -0,0 +1,307 @@
...
+        val r1 = manualRedirect(loginUrl)
+        val r2 = manualRedirect(r1)
+        val loginPageUrl = HonAuthHelpers.absolutize(
+            config.authApiUrl,
+            "$r2&System=IoT_Mobile_App&RegistrationSubChannel=hOn",
+        )
+
+        // Login page (aura): fwuid + loaded must be sent back identical in the POST.
+        val pageText = http.get(loginPageUrl) { ua() }.bodyAsText()
+        val fw = FWUID_RE.find(pageText)!!
```
AGRFesta: 🤖 **Reviewer** — _severity: medium · confidence: high_

`authenticate()`/`refresh()` declare `Either<HonAuthFailure, Unit>` but can throw: `!!` on scraped-page regexes (here and `HREF_RE` in `fetchTokens`), unguarded `json.readTree` on external bodies (`refresh`, `submitLogin`, `apiAuth`), and raw ktor network/timeout exceptions from every `http.get/post`. ARCHITECTURE.md's exception boundary rule ("catch infrastructure exceptions ... convert to typed errors") and the in-repo precedent (`NetatmoClient` wraps these paths in `try/catch` → typed failures) both say these should become `HonLoginFlowBroken`/typed failures. Scraping regexes over a live Salesforce page are exactly the inputs that will eventually miss — an NPE out of a method returning `Either` is the worst version of that day. Gemini flagged the individual `!!`/`readTree` sites; I'd treat it as one sweep over the class, with one caveat on its suggestions: any `try/catch`/`runCatching` around suspend calls must rethrow `CancellationException` (catch `IOException`/`JsonProcessingException` specifically, as NetatmoClient does) — that also applies to the existing `runCatching { json.readTree(resp.bodyAsText()) }` in `HonApiClient.request`, where the suspend `bodyAsText()` belongs outside the block.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonAuth.kt
```diff
@@ -0,0 +1,307 @@
...
+        val href = HREF_RE.find(resp.bodyAsText())!!.groupValues[1]
+        if ("ProgressiveLogin" in href) {
+            val progText = http.get(HonAuthHelpers.absolutize(config.authApiUrl, href)) { ua() }
+                .bodyAsText()
+            // With email 2FA enabled THIS page is the OTP step: unattainable unattended.
+            if (HonAuthHelpers.isProgressiveOtp(progText)) return HonMfaRequired.left()
+        }
+        val tokenPage = http.get(HonAuthHelpers.absolutize(config.authApiUrl, href)) { ua() }
+            .bodyAsText()
+        val tokens = HonAuthHelpers.parseTokenFragment(tokenPage)
```
AGRFesta: 🤖 **Reviewer** — _severity: medium · confidence: high_

`HonOAuthTokens.complete` is computed by the helper, pinned by a helper test — and never read by production code (only tests use it). Both adoption sites (here and the SSO shortcut in `authenticate`) adopt whatever was parsed, and `adopt()` unconditionally persists `tokens.refreshToken`. The sharp edge: on a token-page shape change, `refreshToken` is `""` and the upsert **overwrites the last known-good persisted refresh token with an empty string**, downgrading every future recovery to a full credential login (and see the lockout concern on `reauthenticate`). Check `tokens.complete` at both sites and fail with `HonLoginFlowBroken` before adopting — which also gives the dead `complete` field its reason to exist.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonApiClient.kt
```diff
@@ -0,0 +1,235 @@
...
+    private suspend fun request(
+        exec: suspend () -> HttpResponse,
+    ): Either<HonFailure, JsonNode> {
+        ensureAuthenticated()
+        var resp = exec()
+        if (resp.status.value in RETRY_STATUSES) {
+            reauthenticate()
+            resp = exec()
+        }
+        if (resp.status.value in RETRY_STATUSES) return HonUnauthorized.left()
```
AGRFesta, line 212: 🤖 **Reviewer** — _severity: medium · confidence: medium_

`request()` only treats 401/403 as failures. A 500/429/400 with a JSON body flows through as `Right(errorJson)`; callers then navigate into it — e.g. `loadAttributes` returns `Right(MissingNode)` for a server error, which downstream code can't distinguish from a legitimately empty payload. `loadAppliances` masks it as an empty appliance list. If treating non-2xx JSON as payload is a deliberate port of addhOn behaviour (the cloud reportedly answers 200 with `resultCode` markers), document that in the KDoc; otherwise a typed failure for `status >= 400` (alongside `HonNonJsonResponse`) would make the transport honest about server errors.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonApiClient.kt
```diff
@@ -0,0 +1,235 @@
...
+        return runCatching { json.readTree(resp.bodyAsText()) }
+            .fold({ it.right() }, { HonNonJsonResponse.left() })
+    }
+
+    override fun close() {
```
AGRFesta, line 221: 🤖 **Reviewer** — _severity: low · confidence: high_

Agree with gemini here: `close()` closes `http` but not `auth`, whose two internal `HttpClient`s (and OkHttp thread pools) leak. Impact is low for a singleton whose lifetime is the app's, but the class already opted into `AutoCloseable` — half-closing is the worst of both. Make `HonAuth` closeable and cascade: `auth.close()` here.

providers/src/test/kotlin/org/agrfesta/sh/api/providers/hon/HonAuthTest.kt
```diff
@@ -0,0 +1,396 @@
...
+import io.ktor.http.HttpStatusCode
+import io.ktor.http.headersOf
+import io.mockk.every
+import io.mockk.mockk
+import kotlinx.coroutines.runBlocking
+import org.agrfesta.sh.api.core.application.ports.outbounds.RandomGenerator
+import org.agrfesta.sh.api.core.application.ports.outbounds.TimeProvider
+import org.agrfesta.sh.api.providers.FakePropertyRepository
+import org.agrfesta.sh.api.providers.createMockEngine
+import org.agrfesta.sh.api.providers.netatmo.BehaviorRegistry
```
AGRFesta, line 21: 🤖 **Reviewer** — _severity: low · confidence: medium_

The hOn tests import `BehaviorRegistry`/`ResponseSpec`/`getBodyAsString` from `org.agrfesta.sh.api.providers.netatmo` — shared transport-test infrastructure living in a provider-specific package, now consumed by a second provider. `createMockEngine`/`MockEngineFactory` already sit at the neutral `providers` package level; the registry types belong next to them. Pure test-code organization, fine as a follow-up commit — but the longer it stays, the more imports fossilize.

AGRFesta:
🤖 **Reviewer** — review summary (round 1, sha `d8261921711090fd333ba0a4b7bf4daf440b1fc0`)

Round 1 review of the hOn transport (#247), full diff vs `dev`. Solid overall: the wire-protocol tests are exactly the legitimate use of interaction verification per TDD_WORKFLOW.md, the reverse-engineered quirks are documented and pinned by tests, and the `enabled` gate follows the established provider-gating convention. Main concerns are concurrency (a Mutex the KDoc promises but nobody implements), swallowed typed auth failures that also cause repeated credential submissions to Salesforce (lockout risk), and exceptions escaping the `Either` contract (NetatmoClient, the in-repo precedent, catches and maps them). One convention gap: CLAUDE.md requires a CHANGELOG entry per issue and `## [Unreleased]` has none — an opt-in provider transport behind `providers.hon.enabled` reads like an observable `Added` behaviour (e.g. `- Add hOn (Haier) provider transport: Salesforce OAuth login + authenticated API client, behind providers.hon.enabled. (#247)`); if it is deliberately treated as dark-launch internals until a consumer lands in #248+, say so and I'll drop it. No endpoint was added, so docs/api, API_INDEX and Bruno are correctly untouched.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonApiClient.kt
AGRFesta: ✅ Fixed. Added an `authMutex: Mutex` in `HonApiClient` serializing `ensureAuthenticated`/`forceReauthenticate` (with a double-check inside the lock), so concurrent callers can't consume the same rotated refresh token. The KDoc's promise now matches reality.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonApiClient.kt
AGRFesta: ✅ Fixed. `request()` now maps any non-2xx other than 401/403 to a typed `HonServerError(statusCode)`, so `loadAttributes`/`loadAppliances` no longer pass a server error through as an empty payload. The 200-with-`resultCode` contract (command results) is unchanged and still handled by the callers.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonApiClient.kt
AGRFesta: ✅ Fixed. `HonAuth` now implements `AutoCloseable` (closes both internal `HttpClient`s), and `HonApiClient.close()` cascades `auth.close()`.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonApiClient.kt
AGRFesta: ✅ Fixed, both points. `ensureAuthenticated`/`doReauthenticate` now return `Either<HonFailure, Unit>` and `request()` short-circuits on `Left`, so `HonMfaRequired`/`HonLoginFlowBroken` reach the caller instead of degrading to `HonUnauthorized`. Because the request no longer proceeds after an auth failure, the repeated Salesforce credential submissions are gone; I also removed the redundant `ensureAuthenticated()` in `sendCommand` (post()→request() already authenticates once). Guard test added: an MFA account surfaces `HonMfaRequired` and the login runs exactly once with no appliance call.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonAuth.kt
AGRFesta: ✅ Fixed as one sweep. `authenticate`/`refresh` bodies run inside a `runFlow` boundary that converts `IOException` (ktor network/timeout + Jackson `JsonProcessingException`, which extends it) to `HonLoginFlowBroken`; `CancellationException` is not an `IOException` so it propagates — cooperative cancellation preserved. The `!!` on `FWUID_RE`/`HREF_RE` are now `?: return HonLoginFlowBroken(...).left()`. And per your caveat, the `runCatching` in `HonApiClient.request` now wraps only the non-suspend `json.readTree`, with the suspend `bodyAsText()` moved outside. Guard test added for a login page with no fwuid → `HonLoginFlowBroken`.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonAuth.kt
AGRFesta: ✅ Fixed. `adopt()` now returns `Either` and rejects an incomplete parse with `HonLoginFlowBroken` before touching state — so an empty `refreshToken` can no longer overwrite the persisted good one. Both adoption sites (token page and the SSO shortcut) go through it, which gives `HonOAuthTokens.complete` its production reason to exist. Guard test added: an incomplete token fragment leaves the persisted `HON_REFRESH_TOKEN` untouched.

providers/src/test/kotlin/org/agrfesta/sh/api/providers/hon/HonAuthTest.kt
AGRFesta: Agreed. Left as-is for this PR to keep the hardening diff focused; I'll move `BehaviorRegistry`/`ResponseSpec`/`getBodyAsString` from the `netatmo` package up to the neutral `providers` level (next to `MockEngineFactory`) in a separate test-only follow-up commit before the imports spread further.

AGRFesta:
CHANGELOG: added the `## [Unreleased]` entry — `- Add hOn (Haier) provider transport: Salesforce OAuth login + authenticated API client, behind providers.hon.enabled. (#247)`. Agreed that an opt-in transport behind a gate reads as observable `Added`.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonApiClient.kt
```diff
@@ -0,0 +1,262 @@
...
+     * not hammer Salesforce with repeated logins); a 401/403 surviving the retry becomes
+     * [HonUnauthorized]; any other non-2xx becomes [HonServerError].
+     */
+    // Guard clauses: auth short-circuit, 401 retry, unauthorized, server error, success.
+    @Suppress("ReturnCount")
+    private suspend fun request(
+        exec: suspend () -> HttpResponse,
+    ): Either<HonFailure, JsonNode> {
+        ensureAuthenticated().getOrElse { return it.left() }
+        var resp = exec()
```
AGRFesta, line 207: 🤖 **Reviewer** — _severity: medium · confidence: high_

`exec()` (and the `resp.bodyAsText()` below) can throw raw ktor network exceptions — `ConnectTimeoutException`, `HttpRequestTimeoutException`, `UnknownHostException`, all `IOException`s in ktor 2.3.13 — and nothing between `http.get/post` and the public methods catches them. So `loadAppliances()`/`loadAttributes()`/`sendCommand()` declare `Either<HonFailure, T>` but throw on any network blip, violating the same exception-boundary rule (ARCHITECTURE.md; `NetatmoClient` precedent) that the `HonAuth.runFlow` sweep just fixed. After that sweep this class is the only remaining leak — a mirrored wrapper in `request()` (catch `IOException` → a typed failure, e.g. `HonNetworkError`; `CancellationException` propagates as in `runFlow`) would close the contract for the whole transport. In fairness: this was already true in round 1 and I anchored the finding on `HonAuth` only — flagging it now because the sweep made the asymmetry load-bearing: #248+ consumers will rely on the `Either` being total.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonApiClient.kt
AGRFesta: 🤖 **Reviewer**

Verified: `authMutex` serializes `ensureAuthenticated`/`forceReauthenticate` with a double-check inside the lock, so two concurrent refreshes can no longer consume the same rotated token, and the `HonAuth` KDoc promise is now real. Resolved. One residual noted, not blocking: `mobileId()` still mutates its `var` outside any lock — two concurrent first-ever calls could generate two UUIDs and the last upsert wins. Worst case is a transient mismatch in the device payload on cold start; if you ever see the cloud care about it, move the lazy init under the mutex.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonApiClient.kt
AGRFesta: 🤖 **Reviewer**

Verified: non-2xx other than 401/403 now returns `HonServerError(statusCode)` before any body parsing, and the new test pins a 500 on `loadAttributes` surfacing as `HonServerError(500)` instead of a `Right(MissingNode)` payload. The 200-with-`resultCode` contract is untouched. Resolved.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonApiClient.kt
AGRFesta: 🤖 **Reviewer**

Verified: `HonAuth` implements `AutoCloseable` closing both internal clients, and `HonApiClient.close()` cascades `auth.close()`. Resolved.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonApiClient.kt
AGRFesta: 🤖 **Reviewer**

Verified, both points: `ensureAuthenticated`/`doReauthenticate` return `Either` and `request()` short-circuits on `Left`, so `HonMfaRequired`/`HonLoginFlowBroken` reach the caller instead of degrading to `HonUnauthorized`; the redundant pre-auth in `sendCommand` is gone. The guard test pins exactly the contract I asked for: MFA account → `HonMfaRequired`, login executed once, zero appliance calls (a legitimate non-interaction assertion — the absence IS the domain requirement here). Resolved.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonAuth.kt
AGRFesta: 🤖 **Reviewer**

Verified as a complete sweep: `runFlow` converts `IOException` to `HonLoginFlowBroken` around both flows, the two `!!` regex sites are now guarded early returns (with a guard test for the fwuid-less login page), and in `HonApiClient.request` the suspend `bodyAsText()` moved out of `runCatching` per the caveat. I also checked the one thing that could have silently broken this fix: in the actual ktor 2.3.13 jar, `HttpRequestTimeoutException`, `ConnectTimeoutException` and `SocketTimeoutException` all extend `IOException` (the CancellationException hierarchy is a different-version gotcha), so the catch covers the timeout paths while real cancellation propagates. Resolved. Note: the same Either-totality argument still has one open edge in `HonApiClient.request()` itself — new comment on that line.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonAuth.kt
AGRFesta: 🤖 **Reviewer**

Verified: `adopt()` rejects an incomplete parse with `HonLoginFlowBroken` before touching state or persistence, both adoption sites (token page and SSO shortcut) go through it, and the guard test pins the sharp edge exactly — persisted `HON_REFRESH_TOKEN` survives an incomplete fragment. `complete` now has its production reason to exist. Resolved.

providers/src/test/kotlin/org/agrfesta/sh/api/providers/hon/HonAuthTest.kt
AGRFesta: 🤖 **Reviewer**

Deferral accepted — keeping the hardening diff focused is the right call, and the move is pure test-code organization with no behavioural risk. Resolving on the stated commitment: `BehaviorRegistry`/`ResponseSpec`/`getBodyAsString` move from `netatmo` to the neutral `providers` test package in a test-only follow-up commit before a third consumer fossilizes the imports.

AGRFesta:
🤖 **Reviewer** — review summary (round 2, sha `ae96b634c93250e1c7d31073c637b7352b1c65c4`)

Round 2, delta d826192..ae96b63 verified against each open thread. All six round-1 findings are fixed as claimed and verified in code + guard tests: the promised Mutex now exists (with double-check), auth failures short-circuit request() so HonMfaRequired/HonLoginFlowBroken reach the caller and Salesforce is no longer hammered (guard test pins login-runs-once), non-2xx becomes a typed HonServerError, adopt() rejects incomplete tokens before persisting (guard test pins the persisted token surviving), close() cascades, and the exception boundary holds — I verified in the ktor 2.3.13 jar that HttpRequestTimeoutException/ConnectTimeoutException/SocketTimeoutException all extend IOException, so runFlow's catch covers the ktor timeout paths while CancellationException propagates. CHANGELOG entry added as agreed. One new finding (partly a round-1 miss on my side, now visible by contrast): HonApiClient.request() still lets raw network IOExceptions escape its Either contract, making it the only remaining place where a network blip throws instead of returning a typed HonFailure.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonApiClient.kt
AGRFesta: ✅ Fixed. `request()` now delegates to `requestFlow()` inside a `try/catch (IOException)` boundary that maps raw ktor network exceptions (`ConnectTimeoutException`/`HttpRequestTimeoutException`/`UnknownHostException`, all `IOException` in ktor 2.3.13) to a typed `HonNetworkError` — the mirror of `HonAuth.runFlow`, so `CancellationException` still propagates. This closes the `Either` contract across the whole transport: `loadAppliances`/`loadAttributes`/`sendCommand` no longer throw on a network blip. Guard test added: an engine that answers the auth endpoints but throws `UnknownHostException` on the API call yields `Left(HonNetworkError)`. Fair call that it was latent in round 1 — the `HonAuth` sweep is what made the asymmetry load-bearing.

AGRFesta: 🤖 **Reviewer**

Verified: `request()` wraps `requestFlow()` in a `try/catch (IOException)` -> `HonNetworkError(e.message)` boundary. Coverage checked: both `exec()` invocations (initial + 401 retry) and the suspend `bodyAsText()` live inside `requestFlow`, and every public method (`loadAppliances`, `loadAttributes`, `loadApplianceModel`, `loadCommands`, `sendCommand`) funnels through `get()`/`post()` -> `request()` — no remaining path from a ktor call to a public method escapes the boundary. `CancellationException` extends `RuntimeException`, not `IOException`, so cooperative cancellation still propagates, consistent with `runFlow`. The guard test pins the contract end-to-end: real auth succeeds, the API call throws `UnknownHostException`, and the caller gets a typed `HonNetworkError` instead of a throw. The `Either` contract of the transport is now total. Resolved.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonAuth.kt
AGRFesta: 🤖 **Reviewer**

Closing formally — this was verified in the round-2 summary but the thread never got its verdict reply. Verified: `authenticate`/`refresh` run inside the `runFlow` boundary converting `IOException` (which covers ktor timeouts and Jackson's `JsonProcessingException`) to `HonLoginFlowBroken`, while `CancellationException` propagates; the `!!` on `FWUID_RE`/`HREF_RE` are now guarded early returns; and the `runCatching` in `HonApiClient` wraps only the non-suspend `json.readTree`, with the suspend `bodyAsText()` outside the block. Guard test pins the no-fwuid page -> `HonLoginFlowBroken`. Resolved.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonAuth.kt
AGRFesta: 🤖 **Reviewer**

Closing formally — verified in the round-2 summary but the thread never got its verdict reply. Verified: `adopt()` returns `Either` and rejects an incomplete parse with `HonLoginFlowBroken` before touching any state, so an empty `refreshToken` can no longer overwrite the persisted good one. Both adoption sites go through it, and I checked the SSO shortcut specifically: `adopt(...).getOrElse { return it.left() }` propagates the failure rather than dropping it. `HonOAuthTokens.complete` now has its production reason to exist, and the guard test pins the persisted `HON_REFRESH_TOKEN` surviving an incomplete fragment. Resolved.

providers/src/test/kotlin/org/agrfesta/sh/api/providers/hon/HonAuthTest.kt
AGRFesta: 🤖 **Reviewer**

Conceded — deferring the move of `BehaviorRegistry`/`ResponseSpec`/`getBodyAsString` to the neutral `providers` package as a separate test-only follow-up commit is reasonable; keeping the hardening diff focused was the right call for this PR. Noting for the record that the round-3 delta added one more consumer of the `netatmo` import, so the follow-up is worth doing soon after merge, before the imports fossilize further. Resolved.

AGRFesta:
🤖 **Reviewer** — review summary (round 3, sha `afbd2a74abd11f04338b51d2801bf9abed522de4`)

Round 3, delta ae96b63..afbd2a7. The delta is exactly the fix for the one open finding: `request()` now wraps `requestFlow()` in an `IOException -> HonNetworkError` boundary that covers both `exec()` invocations and the suspend `bodyAsText()`, with `CancellationException` (not an `IOException`) propagating — mirroring `HonAuth.runFlow`. All five public methods route through `get()`/`post()` -> `request()`, so the `Either` contract is now total for the whole transport; the guard test pins a raw `UnknownHostException` surfacing as a typed `HonNetworkError` after a successful auth. No new findings in the delta. Also closing two round-1 threads (runFlow sweep, adopt/incomplete-tokens) that were verified in the round-2 summary but never received their closing verdict reply, and conceding the test-infrastructure placement thread on the author's follow-up commitment. All reviewer threads are now resolved — from the reviewer's side this PR has converged.

PR #255, opened 2026-07-16
issue #248: provider diagnostics endpoint — named read-only probes wi…

### LINKED ISSUE #248: Provider-level diagnostics endpoint (raw probe inspection)

AGRFesta:
…th raw passthrough, hOn implementation


providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonProviderInspector.kt
```diff
@@ -0,0 +1,69 @@
...
+            return MissingProbeParams(probe, expectedParams).left()
+        }
+        return runBlocking {
+            when (probe) {
+                "appliance-list" -> client.rawApplianceList()
+                "context" -> client.rawContext(params.withDefaults("category" to "CYCLE"))
+                "commands" -> client.rawCommands(
+                    params.withDefaults("os" to HonConstants.OS, "appVersion" to HonConstants.APP_VERSION)
+                )
+                else -> client.rawApplianceModel(params)
```
AGRFesta: 🤖 **Reviewer** — _severity: medium · confidence: high_

**Silent misrouting risk: the `else` branch couples the dispatch to the catalog implicitly.** The probe catalog (`requiredParams`) and the dispatch (`when (probe)`) must stay in sync by hand: adding a fifth probe to `requiredParams` without adding its `when` branch compiles fine, passes validation, and silently routes the new probe to `GET /commands/v1/appliance-model` — a wrong-but-200 diagnostic answer, the worst failure mode for a tool whose whole point is showing the provider's truth. Make the `"appliance-model"` branch explicit and turn `else` into a loud failure (e.g. `error("probe '$probe' is in the catalog but has no dispatch branch")`), or restructure so catalog and dispatch cannot drift (e.g. a `Map<String, ProbeSpec>` where the spec carries both the required params and the backing call).

app/src/main/kotlin/org/agrfesta/sh/api/controllers/ProvidersController.kt
```diff
@@ -0,0 +1,79 @@
...
+@RequestMapping("/providers")
+class ProvidersController(
+    private val inspectProviderUseCase: InspectProviderUseCase
+) {
+    /**
+     * ❗ Diagnostic contract, not an API contract: the payload is the provider's raw response,
+     * carries no schema guarantee, and may change or disappear at any time.
+     */
+    @GetMapping("/{provider}/diagnostics")
+    fun diagnostics(
```
AGRFesta, line 33: 🤖 **Reviewer** — _severity: low · confidence: medium_

**Missing integration test for the new flow.** `docs/TDD_WORKFLOW.md`'s test-level table prescribes one integration test per flow covering the happy path. This PR adds unit + MVC slice coverage, but nothing exercises the real wiring: `InspectProviderService`'s optional `Set<InspectableProvider>` injection (Kotlin default + `@ConditionalOnHon`) is precisely the kind of Spring wiring a slice test cannot catch — the slice mocks the use case and the unit test constructs the service by hand. Existing integration tests only prove the context boots with the gate off (empty set). If wiring a mock Ktor engine into the full context for the gate-on path is too heavy for this PR, that's a reasonable call — but then it deserves a stated follow-up rather than silence.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/ports/inbounds/diagnostics/InspectProviderUseCase.kt
```diff
@@ -0,0 +1,28 @@
+package org.agrfesta.sh.api.core.application.ports.inbounds.diagnostics
```
AGRFesta, line 1: 🤖 **Reviewer** — _severity: low · confidence: high_

**`docs/ARCHITECTURE.md` subdomain enumeration is now stale.** Section 1 fixes the inbound-port subpackages to `(areas, devices, sensors, heating, alerts, notifications, settings, home)`; this PR introduces `diagnostics` as a new subpackage on both the inbound and outbound side without updating that enumeration. Project rule: docs describe the current state. Please add `diagnostics` to the list (and, since `docs/SUBDOMAINS.md` presents the application-layer packaging as mirroring the subdomain landscape, consider one line there noting diagnostics is a technical surface, not a business subdomain).

core/src/main/kotlin/org/agrfesta/sh/api/core/application/ports/outbounds/diagnostics/InspectableProvider.kt
```diff
@@ -0,0 +1,36 @@
+package org.agrfesta.sh.api.core.application.ports.outbounds.diagnostics
+
+import arrow.core.Either
+import org.agrfesta.sh.api.core.domain.devices.Provider
+import org.agrfesta.sh.api.core.domain.failures.InspectProviderFailure
+
+/**
+ * Provider-level diagnostics capability: runs a named **read-only** probe against the provider's
+ * cloud and returns the raw response body as a JSON [String].
```
AGRFesta, line 9: 🤖 **Reviewer** — _severity: low · confidence: medium_

**The always-JSON guarantee is an endpoint promise enforced only inside one adapter.** `docs/api/providers.md` documents that a non-JSON 200 body is wrapped in a `{"nonJsonBody": ...}` envelope, but that behaviour lives in `HonProviderInspector.asJsonPayload()` — the port KDoc only says the return is "a JSON [String]" without saying what an implementation must do when the cloud answers non-JSON. A future Netatmo/SwitchBot inspector author reading the port contract has no signal to replicate the envelope, and `:core` cannot centralize it (no Jackson). At minimum, state the envelope rule explicitly in this KDoc as part of the port contract, so the next implementation doesn't silently break the documented endpoint behaviour.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/ports/outbounds/diagnostics/InspectableProvider.kt
```diff
@@ -0,0 +1,36 @@
...
+ * ❗ **Read-only, forever**: a probe must never send commands or mutate provider state.
+ * Implementations violating this break the port contract.
+ */
+interface InspectableProvider {
+
+    /** The provider this diagnostics implementation belongs to. */
+    val provider: Provider
+
+    /** The names of the probes this implementation supports. */
+    val probes: Set<String>
```
AGRFesta, line 25: 🤖 **Reviewer** — _severity: low · confidence: medium_

**`probes` has no consumer.** Neither `InspectProviderService` nor the controller reads it; the availability hint reaches the caller via `UnknownProbe.availableProbes`, and the hOn implementation could build that from its private catalog without the port declaring anything. Unused port surface is contract you maintain for free — either drop it from the port, or make it earn its place (e.g. move the unknown-probe check into `InspectProviderService` using `probes`, so every future adapter gets that validation without re-implementing it — which would also pair well with hoisting the self-description duty out of each adapter).

core/src/main/kotlin/org/agrfesta/sh/api/core/domain/failures/Providers.kt
```diff
@@ -0,0 +1,24 @@
...
+) : InspectProviderFailure
+
+/** A required parameter of the probe is missing; [expectedParams] is the full required set. */
+data class MissingProbeParams(
+    val probe: String,
+    val expectedParams: Set<String>
+) : InspectProviderFailure
+
+/** The provider was reached but failed or is unreachable; [message] surfaces the cause. */
+data class ProviderInspectionFailure(val message: String?) : InspectProviderFailure
```
AGRFesta, line 29: 🤖 **Reviewer** — _severity: low · confidence: low_

**502 body text is Kotlin data-class `toString()` of internal transport types.** The adapter maps every `HonFailure` via `ProviderInspectionFailure(it.toString())`, so messages like `HonServerError(statusCode=500)` — internal `:providers` type names — become observable API text, and both the inspector test and the MVC slice test pin them. A rename of a `HonFailure` variant now silently changes the endpoint's output. On a documented no-schema-guarantee surface this is defensible (and I note the sealed `InspectProviderFailure` works as an outbound-port contract only because adapters flatten into this stringly case — a deliberate deviation from the non-sealed port-failure pattern in `docs/ARCHITECTURE.md` §"Two kinds of failure interfaces"). Flagging so the choice is on record; if intended, consider a one-line KDoc note that the message is diagnostic text with no stability guarantee.

docs/api/providers.md
```diff
@@ -0,0 +1,100 @@
...
+### Probes — hOn
+
+| Probe             | Required params                                        | Backing call                                  |
+|-------------------|--------------------------------------------------------|-----------------------------------------------|
+| `appliance-list`  | —                                                      | `POST /unified-api/v1/view/appliance-list`    |
+| `context`         | `macAddress`, `applianceType`                          | `GET /commands/v1/context` (`category` defaults to `CYCLE`) |
+| `commands`        | `macAddress`, `applianceType`, `applianceModelId`, `code` | `GET /commands/v1/retrieve` (`os`/`appVersion` defaulted; optional `firmwareId`, `fwVersion`, `series` passed through) |
+| `appliance-model` | `macAddress`, `code`                                   | `GET /commands/v1/appliance-model`            |
+
+SwitchBot and Netatmo do not expose provider-level probes yet (see the follow-up issue in the
```
AGRFesta: 🤖 **Reviewer** — _severity: low · confidence: low_

**Permanent doc points at an ephemeral artifact.** "see the follow-up issue in the project root" references a temporary file that will be deleted once the hOn epic closes, leaving a dangling pointer in the API docs. Reference the GitHub issue number instead (or drop the parenthetical — the sentence stands on its own).

AGRFesta:
🤖 **Reviewer** — review summary (round 1, sha `5e9377f178c86c74e3cbf6c722a4213489fa67fb`)

Round 1 review of the provider diagnostics endpoint (#248). Solid PR: the port/adapter split is clean, HTTP status mapping mirrors the device-level diagnostics endpoint exactly, the HonApiClient raw/typed refactor preserves the CancellationException boundary, tests follow the Chicago-school conventions (fakes + state at the service, wire-protocol verification only at the transport), and docs/Bruno/changelog are all present. Findings are mostly design-level: a dispatch `else` that can silently misroute future probes, a missing integration test for the new flow, and a few doc/port-contract alignment points. No blocking correctness bug found.

app/src/main/kotlin/org/agrfesta/sh/api/controllers/ProvidersController.kt
AGRFesta: 🤖 **Author** — Agreed on the gap, deferring deliberately. The wiring risk splits in two: (a) the optional `Set<InspectableProvider>` injection with the gate **off** (empty set + Kotlin default) is exercised by every existing integration test booting the full context; (b) the gate-**on** path needs a context flavor with hOn enabled and a stubbed Ktor engine, which would break the `AbstractIntegrationTest` context cache for a diagnostics-only surface. That cost fits the upcoming hOn driver sub-issue of the epic, which needs a gate-on integration context anyway — the provider-diagnostics happy path will ride on it. Tracked in the retrofit follow-up issue.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/ports/outbounds/diagnostics/InspectableProvider.kt
AGRFesta: 🤖 **Author** — Keeping `probes` for this PR: the port shape is the contract approved in the #248 issue spec, and `probes` is what makes the port self-describing rather than a bag of behavior. The suggested hoisting (unknown-probe validation in `InspectProviderService` driven by `probes`) is the direction I'd take too, but it changes tested behavior and only pays off when a second inspector exists. Tracked in the retrofit follow-up issue, where the dedup becomes real.

.env.example
```diff
@@ -28,3 +28,7 @@ PROVIDERS_NETATMO_CLIENTID=
 PROVIDERS_NETATMO_CLIENTSECRET=
 PROVIDERS_NETATMO_HOMEID=
 PROVIDERS_NETATMO_ROOMID=
+
+# hOn (Haier) provider (optional — leave empty to disable)
```
AGRFesta, line 32: 🤖 **Reviewer** — _severity: low · confidence: low_

**"Leave empty to disable" is not how the gate works.** The gate is `providers.hon.enabled` (`@ConditionalOnHon`), and `docker-compose.yml` hardcodes `PROVIDERS_HON_ENABLED=true` — so leaving these blank does *not* disable hOn; it boots the provider with empty credentials and every probe fails at runtime with an auth 502 instead of a clean `provider not enabled` response. The comment faithfully copies the SwitchBot/Netatmo lines above, so this is a pre-existing pattern (all three are gated by hardcoded `ENABLED=true` in compose), not something this PR introduced — flagging it because the new lines replicate it onto a provider whose enabled-gate was a deliberate design decision. Consider `# hOn (Haier) provider (enabled in docker-compose — credentials required)` here, or a follow-up sweep fixing all three comments.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/ports/inbounds/diagnostics/InspectProviderUseCase.kt
AGRFesta: 🤖 **Reviewer**

Verified: `diagnostics` added to the inbound-port enumeration in `docs/ARCHITECTURE.md`, and `docs/SUBDOMAINS.md` both updates the packaging enumeration and adds the note that diagnostics is a technical inspection surface (solution space), not a subdomain. Both suggestions addressed.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/ports/outbounds/diagnostics/InspectableProvider.kt
AGRFesta: 🤖 **Reviewer**

Verified: the port KDoc now states the contract explicitly — verbatim body when valid JSON, `{"nonJsonBody": ...}` envelope for non-JSON 2xx bodies — so a future inspector author gets the rule from the port, not from reverse-engineering the hOn adapter.

core/src/main/kotlin/org/agrfesta/sh/api/core/domain/failures/Providers.kt
AGRFesta: 🤖 **Reviewer**

Verified: `ProviderInspectionFailure` KDoc now states `message` is diagnostic text with no stability guarantee and must never be parsed. The deliberate choice is on record at the type that carries it.

docs/api/providers.md
AGRFesta: 🤖 **Reviewer**

Verified: the ephemeral repo-root file reference is gone; the doc now says the retrofit is tracked as a follow-up issue, which survives the epic's cleanup.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonProviderInspector.kt
AGRFesta: 🤖 **Reviewer**

Verified: `"appliance-model"` is now an explicit branch and the `else` fails loudly with `error(...)` plus a comment stating the invariant. A catalog/dispatch drift can no longer silently misroute to a wrong-but-200 probe answer — it now blows up on first use, which is the correct failure mode for a diagnostic surface. The guard being untested is fine: it is unreachable by construction, defensive programming rather than behaviour.

app/src/main/kotlin/org/agrfesta/sh/api/controllers/ProvidersController.kt
AGRFesta: 🤖 **Reviewer**

Conceding. The rejection does exactly what the original comment asked for: the deferral is stated and tracked, the gate-off wiring (empty `Set<InspectableProvider>` + Kotlin default) is genuinely covered by every existing full-context test, and breaking the `AbstractIntegrationTest` context cache for a diagnostics-only surface is a real cost that the hOn driver sub-issue will have to pay anyway. Riding the provider-diagnostics happy path on that context is the right sequencing.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/ports/outbounds/diagnostics/InspectableProvider.kt
AGRFesta: 🤖 **Reviewer**

Conceding. `probes` being part of the issue-spec-approved port shape is a fair authority argument, and the hoisting direction (unknown-probe validation in `InspectProviderService` driven by `probes`) is agreed and tracked for when a second inspector makes the dedup real. Keeping self-description on the port until then is defensible.

AGRFesta:
🤖 **Reviewer** — review summary (round 2, sha `fdc99ce45b9b3d48e9363e1b91254a9b459940ce`)

Round 2 re-review, scoped to the delta since `5e9377f` plus the open threads. All 7 round-1 threads are closed: five were fixed in code exactly as requested (docs subdomain enumeration, port KDoc envelope contract, failure-message stability note, ephemeral doc pointer, and the dispatch `else` turned into an explicit branch plus a loud guard), and two author rejections (integration test deferral, keeping `probes` on the port) are conceded — both are justified deferrals with a tracked follow-up. One new low-severity finding in the delta: the `.env.example` comment for hOn claims empty credentials disable the provider, but the gate is `providers.hon.enabled`, which docker-compose hardcodes to `true`. No blocking issues; the PR converges.

PR #256, opened 2026-07-17
issue #254: retrofit provider-level diagnostics probes to SwitchBot (…

### LINKED ISSUE #254: Retrofit provider-level diagnostics probes to SwitchBot and Netatmo
# Retrofit provider-level diagnostics probes to SwitchBot and Netatmo

Follow-up of #248, which introduced `GET /providers/{provider}/diagnostics` (the
`InspectableProvider` outbound port + `InspectProviderService` use case) with an hOn-only
implementation. The issue explicitly allowed deferring the SwitchBot/Netatmo retrofit to keep the
epic moving — this is that deferred work.

## Goal

Register an `InspectableProvider` implementation for SwitchBot and Netatmo, so the existing
endpoint answers for them too (today they get `501 Not Implemented`). No endpoint, use-case or port
changes: the routing, failure taxonomy (`UnknownProbe`, `MissingProbeParams`,
`ProviderInspectionFailure`) and HTTP mapping are already in place and tested.

### SwitchBot (`@ConditionalOnSwitchBot`)

| Probe     | Params | Backing call                     |
|-----------|--------|----------------------------------|
| `devices` | —      | `GET /devices` (account devices) |

`SwitchBotDevicesClient` already has the raw-passthrough precedent (`getDeviceStatusRaw`); add the
raw variant of the devices list call and wire a `SwitchBotProviderInspector`.

### Netatmo (`@ConditionalOnNetatmo`)

| Probe         | Params   | Backing call                    |
|---------------|----------|---------------------------------|
| `home-status` | `homeId` | `GET /api/homestatus?home_id=…` |

Return the body verbatim (no room scoping — that is the device-level diagnostics' job).

## Guardrails (same as #248 — see `docs/api/providers.md`)

- Read-only probes only; never send commands or mutate provider state.
- Raw passthrough, no schema guarantee: diagnostic contract, not an API contract.
- No tokens/credentials/auth headers in responses or logs.
- Requires the provider's conditional gate to be enabled.

## Testing

Mirror `HonProviderInspectorTest`: slice test per provider against the stubbed HTTP engine —
unknown probe → available probes hint, param validation (Netatmo), raw body passthrough, transport
failure → `ProviderInspectionFailure`. No use-case or controller changes to test.

## DoD

`./gradlew build` green (incl. detekt). Update the probe tables in `docs/api/providers.md` (and
remove its "not yet" note). CHANGELOG (`Added`): SwitchBot/Netatmo provider diagnostics probes.

AGRFesta:
…devices) and Netatmo (home-status)

### DIFF
--- a/docs/api/providers.md
+++ b/docs/api/providers.md
@@ -31,7 +31,7 @@ any environment, without deploying ad-hoc code.
 | Param   | Example          | Description                                                    |
 |---------|------------------|----------------------------------------------------------------|
 | `probe` | `appliance-list` | **Required.** The probe to run.                                |
-| *(any)* | `macAddress=...` | Probe-specific parameters, passed through to the backing call. |
+| *(any)* | `macAddress=...` | Probe-specific parameters. hOn probes forward any extra params to the backing call; SwitchBot and Netatmo probes use only their listed params. |
 
 ### Probes — hOn
 
@@ -42,8 +42,22 @@ any environment, without deploying ad-hoc code.
 | `commands`        | `macAddress`, `applianceType`, `applianceModelId`, `code` | `GET /commands/v1/retrieve` (`os`/`appVersion` defaulted; optional `firmwareId`, `fwVersion`, `series` passed through) |
 | `appliance-model` | `macAddress`, `code`                                   | `GET /commands/v1/appliance-model`            |
 
-SwitchBot and Netatmo do not expose provider-level probes yet — a retrofit is tracked as a
-follow-up issue; their device-level diagnostics remain available via `GET /devices/{uuid}/diagnostics`.
+### Probes — SwitchBot
+
+| Probe     | Required params | Backing call                     |
+|-----------|-----------------|----------------------------------|
+| `devices` | —               | `GET /devices` (account devices) |
+
+### Probes — Netatmo
+
+| Probe         | Required params | Backing call                    |
+|---------------|-----------------|---------------------------------|
+| `home-status` | `homeId`        | `GET /api/homestatus?home_id=…` |
+
+The body is returned verbatim, with no room scoping — scoping to the configured room is the
+device-level diagnostics' job (`GET /devices/{uuid}/diagnostics`).
+`homeId` is always explicit — the probe never falls back to the configured home, so what you
+probe is always what you asked for.
 
 ### Walkthrough — querying hOn step by step
 
--- a/providers/src/main/kotlin/org/agrfesta/sh/api/providers/netatmo/NetatmoProviderInspector.kt
+++ b/providers/src/main/kotlin/org/agrfesta/sh/api/providers/netatmo/NetatmoProviderInspector.kt
@@ -1,0 +1,42 @@
+package org.agrfesta.sh.api.providers.netatmo
+
+import arrow.core.Either
+import arrow.core.flatMap
+import arrow.core.left
+import kotlinx.coroutines.runBlocking
+import org.agrfesta.sh.api.core.application.ports.outbounds.diagnostics.InspectableProvider
+import org.agrfesta.sh.api.core.domain.devices.Provider
+import org.agrfesta.sh.api.core.domain.failures.InspectProviderFailure
+import org.agrfesta.sh.api.core.domain.failures.MissingProbeParams
+import org.agrfesta.sh.api.core.domain.failures.ProviderInspectionFailure
+import org.agrfesta.sh.api.core.domain.failures.UnknownProbe
+import org.agrfesta.sh.api.providers.asJsonPayload
+import org.springframework.stereotype.Service
+
+/**
+ * Provider-level diagnostics for Netatmo: read-only probes backed by [NetatmoClient], returning
+ * the cloud's response body **verbatim** — no parse → re-serialize round trip, no room scoping —
+ * so the real shapes can be re-captured in any environment.
+ */
+@Service
+@ConditionalOnNetatmo
+class NetatmoProviderInspector(
+    private val client: NetatmoClient
+) : InspectableProvider {
+
+    override val provider: Provider = Provider.NETATMO
+
+    override val probes: Set<String> = setOf("home-status")
+
+    override fun inspect(probe: String, params: Map<String, String>): Either<InspectProviderFailure, String> {
+        if (probe !in probes) return UnknownProbe(probe, probes).left()
+        val homeId = params["homeId"] ?: return MissingProbeParams(probe, setOf("homeId")).left()
+        return Either.catch { runBlocking { client.getHomeStatusRaw(homeId) } }
+            .mapLeft { ProviderInspectionFailure(it.toString()) }
+            .flatMap { outcome ->
+                outcome
+                    .map { it.asJsonPayload() }
+                    .mapLeft { ProviderInspectionFailure(it.toString()) }
+            }
+    }
+}
--- a/providers/src/main/kotlin/org/agrfesta/sh/api/providers/switchbot/SwitchBotDevicesClient.kt
+++ b/providers/src/main/kotlin/org/agrfesta/sh/api/providers/switchbot/SwitchBotDevicesClient.kt
@@ -36,12 +36,16 @@ class SwitchBotDevicesClient(
         }
     }
 
-    suspend fun getDevices(): JsonNode {
-        val content = client.get("${config.baseUrl}/devices") {
+    suspend fun getDevices(): JsonNode = mapper.readTree(getDevicesRaw())
+
+    /**
+     * Returns the account devices response body exactly as the provider sent it, with no parsing,
+     * so it can be passed through verbatim as the diagnostics payload.
+     */
+    suspend fun getDevicesRaw(): String =
+        client.get("${config.baseUrl}/devices") {
             switchbotHeaders()
         }.bodyAsText()
-        return mapper.readTree(content)
-    }
 
     suspend fun getDeviceStatus(deviceId: String): JsonNode =
         mapper.readTree(getDeviceStatusRaw(deviceId))
--- a/providers/src/test/kotlin/org/agrfesta/sh/api/providers/switchbot/SwitchBotProviderInspectorTest.kt
+++ b/providers/src/test/kotlin/org/agrfesta/sh/api/providers/switchbot/SwitchBotProviderInspectorTest.kt
@@ -1,0 +1,107 @@
+package org.agrfesta.sh.api.providers.switchbot
+
+import com.fasterxml.jackson.databind.ObjectMapper
+import io.kotest.assertions.arrow.core.shouldBeLeft
+import io.kotest.assertions.arrow.core.shouldBeRight
+import io.kotest.assertions.withClue
+import io.kotest.matchers.shouldBe
+import io.kotest.matchers.string.shouldContain
+import io.kotest.matchers.types.shouldBeInstanceOf
+import io.ktor.http.HttpMethod
+import io.ktor.http.HttpStatusCode
+import io.mockk.every
+import io.mockk.mockk
+import org.agrfesta.sh.api.core.application.ports.outbounds.RandomGenerator
+import org.agrfesta.sh.api.core.application.ports.outbounds.TimeProvider
+import org.agrfesta.sh.api.core.domain.failures.ProviderInspectionFailure
+import org.agrfesta.sh.api.core.domain.failures.UnknownProbe
+import org.agrfesta.sh.api.providers.createMockEngine
+import org.agrfesta.sh.api.providers.netatmo.BehaviorRegistry
+import org.agrfesta.sh.api.providers.netatmo.ResponseSpec
+import org.junit.jupiter.api.Test
+import java.time.Instant
+
+class SwitchBotProviderInspectorTest {
+    private val config = SwitchBotConfiguration(
+        baseUrl = "https://api.switchbot.example.com",
+        token = "sb-token",
+        secret = "sb-secret",
+    )
+    private val timeProvider: TimeProvider = mockk {
+        every { now() } returns Instant.parse("2026-07-17T10:00:00Z")
+    }
+    private val randomGenerator: RandomGenerator = mockk {
+        every { string() } returns "fixed-nonce"
+    }
+    private val registry = BehaviorRegistry()
+    private val engine = createMockEngine(registry)
+    private val client = SwitchBotDevicesClient(config, ObjectMapper(), timeProvider, randomGenerator, engine)
+
+    private val sut = SwitchBotProviderInspector(client)
+
+    @Test
+    fun `inspect() with an unknown probe returns UnknownProbe carrying the available probes`() {
+        // Given
+        val unknownProbe = "send-command"
+
+        // When
+        val result = sut.inspect(unknownProbe, emptyMap()).shouldBeLeft()
+
+        // Then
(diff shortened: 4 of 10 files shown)


providers/src/main/kotlin/org/agrfesta/sh/api/providers/netatmo/NetatmoProviderInspector.kt
```diff
@@ -0,0 +1,43 @@
...
+) : InspectableProvider {
+
+    override val provider: Provider = Provider.NETATMO
+
+    override val probes: Set<String> = setOf("home-status")
+
+    override fun inspect(probe: String, params: Map<String, String>): Either<InspectProviderFailure, String> {
+        if (probe !in probes) return UnknownProbe(probe, probes).left()
+        val homeId = params["homeId"] ?: return MissingProbeParams(probe, setOf("homeId")).left()
+        return Either.catch { runBlocking { client.getHomeStatusRaw(homeId) } }
```
AGRFesta, line 34: 🤖 **Reviewer** — _severity: medium · confidence: medium_

**Doc contract mismatch — probe params are not passed through.** `docs/api/providers.md` (Query parameters table) promises: "*(any)* — Probe-specific parameters, passed through to the backing call", and hOn honors it (`client.rawContext(params.withDefaults(...))` forwards the whole map). This inspector extracts only `homeId` and silently drops everything else — e.g. Netatmo's `homestatus` accepts a `device_types` filter a caller could legitimately expect to reach the backing call. Either forward the remaining params (hOn precedent), or scope the doc's pass-through claim to the probes that actually honor it. As-is the endpoint's self-describing promise is broken for Netatmo.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/switchbot/SwitchBotDevicesClient.kt
```diff
@@ -43,6 +43,15 @@ class SwitchBotDevicesClient(
         return mapper.readTree(content)
     }
 
+    /**
+     * Returns the account devices response body exactly as the provider sent it, with no parsing,
+     * so it can be passed through verbatim as the diagnostics payload.
+     */
+    suspend fun getDevicesRaw(): String =
```
AGRFesta, line 45: 🤖 **Reviewer** — _severity: low · confidence: high_

**Duplicated wire call — this file already established the delegation pattern.** `getDeviceStatus` is defined as `mapper.readTree(getDeviceStatusRaw(deviceId))`: raw is the primitive, typed delegates to it. The new `getDevicesRaw` instead copy-pastes the `client.get("${config.baseUrl}/devices") { switchbotHeaders() }.bodyAsText()` block from `getDevices`, so the `/devices` URL and headers now live in two places. Refactor `getDevices()` to `mapper.readTree(getDevicesRaw())` — behavior-neutral, and consistent with the pattern three lines below.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/netatmo/NetatmoProviderInspector.kt
```diff
@@ -0,0 +1,43 @@
...
+    private val client: NetatmoClient
+) : InspectableProvider {
+
+    override val provider: Provider = Provider.NETATMO
+
+    override val probes: Set<String> = setOf("home-status")
+
+    override fun inspect(probe: String, params: Map<String, String>): Either<InspectProviderFailure, String> {
+        if (probe !in probes) return UnknownProbe(probe, probes).left()
+        val homeId = params["homeId"] ?: return MissingProbeParams(probe, setOf("homeId")).left()
```
AGRFesta, line 33: 🤖 **Reviewer** — _severity: low · confidence: medium_

**Consider defaulting `homeId` to the configured one.** `NetatmoConfiguration` already carries `homeId`, and hOn set the precedent of `withDefaults(...)` for params the caller usually doesn't need to supply (with caller override). Defaulting here would make `home-status` a zero-param probe — matching the doc's "when in doubt just call it" self-describing spirit — while still allowing an explicit `homeId` to probe other homes. If requiring it is deliberate (never silently probe the configured home), that's defensible too, but then the choice deserves a line in the probe doc.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/netatmo/NetatmoProviderInspector.kt
```diff
@@ -0,0 +1,43 @@
...
+        return Either.catch { runBlocking { client.getHomeStatusRaw(homeId) } }
+            .fold(
```
AGRFesta: 🤖 **Reviewer** — _severity: low · confidence: medium_

**Nested-Either fold could be flattened.** `Either.catch { ... }` yields `Either<Throwable, Either<NetatmoClientFailure, String>>` and the `fold` re-implements what `mapLeft` + `flatMap` say directly:

```kotlin
return Either.catch { runBlocking { client.getHomeStatusRaw(homeId) } }
    .mapLeft { ProviderInspectionFailure(it.toString()) }
    .flatMap { outcome ->
        outcome
            .map { it.asJsonPayload() }
            .mapLeft { ProviderInspectionFailure(it.toString()) }
    }
```

Pure style (ARCHITECTURE.md: chain `Either` with Arrow operators); take it or leave it.

docs/api/providers.md
```diff
@@ -42,8 +42,20 @@ any environment, without deploying ad-hoc code.
...
+
+| Probe     | Required params | Backing call                     |
+|-----------|-----------------|----------------------------------|
+| `devices` | —               | `GET /devices` (account devices) |
+
+### Probes — Netatmo
+
+| Probe         | Required params | Backing call                    |
+|---------------|-----------------|---------------------------------|
+| `home-status` | `homeId`        | `GET /api/homestatus?home_id=…` |
```
AGRFesta, line 55: 🤖 **Reviewer** — _severity: low · confidence: medium_

**Bruno collection not updated for the new probes.** CLAUDE.md's convention covers modified endpoints too, and this same doc promises (Walkthrough section): "The Bruno collection (`bruno/providers/`) has a ready-made request with all the optional params pre-wired as disabled entries." `Get Provider Diagnostics.bru` is still hOn-only (`provider: hon`, hOn params) — a `~homeId` disabled entry (and perhaps a comment naming the SwitchBot/Netatmo probe names) would keep that promise true. Related doc-freshness nit: the `501` example further down still uses `'NETATMO'` as its subject, which now reads as "Netatmo has no diagnostics" — a provider that genuinely lacks an implementation would be a less confusing example.

providers/src/test/kotlin/org/agrfesta/sh/api/providers/switchbot/SwitchBotProviderInspectorTest.kt
```diff
@@ -0,0 +1,107 @@
...
+import io.ktor.http.HttpMethod
+import io.ktor.http.HttpStatusCode
+import io.mockk.every
+import io.mockk.mockk
+import org.agrfesta.sh.api.core.application.ports.outbounds.RandomGenerator
+import org.agrfesta.sh.api.core.application.ports.outbounds.TimeProvider
+import org.agrfesta.sh.api.core.domain.failures.ProviderInspectionFailure
+import org.agrfesta.sh.api.core.domain.failures.UnknownProbe
+import org.agrfesta.sh.api.providers.createMockEngine
+import org.agrfesta.sh.api.providers.netatmo.BehaviorRegistry
```
AGRFesta, line 19: 🤖 **Reviewer** — _severity: low · confidence: low_

**Cross-provider test fixture living in the `netatmo` package.** `BehaviorRegistry`/`ResponseSpec` sit in `testFixtures/.../providers/netatmo/` but are now imported by hOn, SwitchBot and Netatmo tests alike — this PR adds the third consumer. Pre-existing smell, not introduced here, but each new import strengthens the case for hoisting them one level up to `org.agrfesta.sh.api.providers` (next to `createMockEngine`, which already lives there). Fine as a follow-up refactor issue rather than in this PR.

AGRFesta:
🤖 **Reviewer** — review summary (round 1, sha `f27842c4f86d73be969fbc5b0f534dd1565ca2c4`)

Round 1 review of the SwitchBot/Netatmo diagnostics retrofit (#254). Overall a clean retrofit: both inspectors follow the hOn precedent, the shared `asJsonPayload` extraction removes duplication instead of adding a third copy, tests are Chicago-style against the mock engine with wire-protocol verification only where the interaction is the contract (homeId forwarding), and changelog + API doc are updated. Findings below are consistency/refactor level; the one medium item is a doc-vs-implementation mismatch on probe param pass-through.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/switchbot/SwitchBotDevicesClient.kt
AGRFesta: @

🤖 **Author** — accepted

`getDevices()` now delegates: `mapper.readTree(getDevicesRaw())` — the `/devices` URL and headers live in one place, consistent with the `getDeviceStatus` pattern three lines below.
@

providers/src/main/kotlin/org/agrfesta/sh/api/providers/netatmo/NetatmoProviderInspector.kt
AGRFesta: 🤖 **Author** — accepted (doc scoping)

Fixed by scoping the doc's pass-through claim: the *(any)* row now states that hOn probes forward extra params to the backing call while SwitchBot and Netatmo probes use only their listed params. Chose the doc fix over real forwarding because #254 defines `home-status` with `homeId` only; Netatmo pass-through (e.g. `device_types`) can come with its own tested cycle if a probe ever needs it.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/netatmo/NetatmoProviderInspector.kt
AGRFesta: 🤖 **Author** — accepted (doc path)

Kept `homeId` required, as specified in #254: the `withDefaults` precedent in hOn covers protocol constants (`os`, `appVersion`, `category`), not account-scoped identifiers, and on a diagnostic surface an explicit target beats a silent fallback. Documented the choice in the Netatmo probes section.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/netatmo/NetatmoProviderInspector.kt
AGRFesta: 🤖 **Author** — accepted

Replaced the nested fold with `mapLeft` + `flatMap` as suggested.

docs/api/providers.md
AGRFesta: 🤖 **Author** — accepted (Bruno); 501 example kept

Bruno updated with a `~homeId` disabled entry and a docs block listing the probe catalog per provider. Kept `'NETATMO'` in the 501 example: since #254 every provider has an implementation, the only real 501 is the disabled-gate case — which the sentence above the example already states; swapping the provider name wouldn't change that.

providers/src/test/kotlin/org/agrfesta/sh/api/providers/switchbot/SwitchBotProviderInspectorTest.kt
AGRFesta: 🤖 **Author** — deferred (follow-up)

Agreed on the smell — `BehaviorRegistry`/`ResponseSpec` belong next to `createMockEngine` in the root `providers` fixture package. Out of scope here as you note; tracked as a follow-up refactor issue.

docs/api/providers.md
AGRFesta: 🤖 **Reviewer**

Verified: `Get Provider Diagnostics.bru` now has the `~homeId` disabled entry and a docs block listing the probe catalog per provider — the walkthrough's promise holds again. Conceded on the 501 example: since #254 every provider ships an implementation, the disabled-gate sentence directly above the example already carries the explanation, so swapping the provider name would add nothing.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/netatmo/NetatmoProviderInspector.kt
AGRFesta: 🤖 **Reviewer**

Verified: the doc path was taken — the Netatmo probes section now states that `homeId` is always explicit and never falls back to the configured home. The distinction drawn (hOn's `withDefaults` covers protocol constants, not account-scoped identifiers) is a sharper rule than my suggestion; explicit target on a diagnostic surface is the right call.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/netatmo/NetatmoProviderInspector.kt
AGRFesta: 🤖 **Reviewer**

Verified: the *(any)* row now scopes the pass-through claim accurately — hOn forwards extra params, SwitchBot and Netatmo use only their listed params. Doc-fix over real forwarding is reasonable given #254 defines `home-status` with `homeId` only; if a Netatmo probe ever needs `device_types`, it gets its own tested cycle.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/netatmo/NetatmoProviderInspector.kt
AGRFesta: 🤖 **Reviewer**

Verified: the nested fold is now `mapLeft` + `flatMap`, exactly as sketched; `NetatmoProviderInspectorTest` is green.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/switchbot/SwitchBotDevicesClient.kt
AGRFesta: 🤖 **Reviewer**

Verified: `getDevices()` now delegates to `mapper.readTree(getDevicesRaw())` — the `/devices` URL and headers live in one place, and the SwitchBot tests pass.

providers/src/test/kotlin/org/agrfesta/sh/api/providers/switchbot/SwitchBotProviderInspectorTest.kt
AGRFesta: 🤖 **Reviewer**

Agreed — deferring the `BehaviorRegistry`/`ResponseSpec` hoist to a dedicated refactor issue is exactly what I proposed. Closing this thread; the follow-up issue is the tracker.

AGRFesta:
🤖 **Reviewer** — review summary (round 2, sha `4cdec721e4d905fab5186899d17f6cbd17c0c71b`)

Round 2: verified the delta f27842c..4cdec72 against all six round-1 threads. Every accepted fix landed as described (Bruno probe catalog + ~homeId entry, doc scoping of the pass-through claim, explicit-homeId rationale in the Netatmo probes section, mapLeft+flatMap flattening, getDevices delegation to getDevicesRaw), the two rejections/deferrals are well-argued, and the affected provider tests are green. No new findings in the delta — all threads closed.

Issue #251, opened 2026-07-16
AC control: `AirConditioner` outbound port + hOn write path


AGRFesta:
# hOn AC — write path (commands) + raw-state inspection (#251)

> **Note (revised after #250 findings).** #250 was closed: live device captures proved the
> AC exposes no usable ambient readings (no humidity; indoor/outdoor temperatures are
> position/coil-biased and untrustworthy), so the AC is **not a sensor**. Its real role is an
> **actuator**. #250's only surviving deliverable — `Inspectable.inspect()` for raw AC state —
> is folded into this issue, where raw-state visibility is needed while implementing writes.

## Goal

The domain can command an air conditioner — power, mode, target temperature, fan speed —
and the hOn driver implements it against the cloud. Scope agreed for v1: **on/off + mode +
target temperature + fan speed**; swing/eco/sleep/silent are follow-ups.

Additionally, the hOn AC driver exposes its raw cloud state via `Inspectable` (carried over
from #250) for diagnostics and to ease debugging the write path.

### Core (domain + port)

- New outbound port in `core/.../ports/outbounds/devices/`:

  ```kotlin
  interface AirConditioner : OnOffActuator {
      fun setMode(mode: AcMode): Either<ActuatorOperationFailure, Unit>
      fun setTargetTemperature(temperature: BigDecimal): Either<ActuatorOperationFailure, Unit>
      fun setFanSpeed(speed: AcFanSpeed): Either<ActuatorOperationFailure, Unit>
  }
  ```

  with domain enums `AcMode` (`AUTO, COOL, HEAT, DRY, FAN_ONLY`) and `AcFanSpeed`
  (`AUTO, LOW, MEDIUM, HIGH`). Exact shape (single `applySettings(...)` vs per-property
  methods) to be settled in the Phase 0 test-list discussion — note that hOn's `settings`
  command **always transmits all parameters**, which argues for a read-modify-write inside
  the driver either way.
- Update the AC `DevicePrototype` roles if a dedicated feature/role is needed for the
  use-case layer to find controllable ACs (reuse `ACTUATOR` if sufficient). `HonAc` implements
  `AirConditioner` (actuator) **and** `Inspectable`; no `SENSOR` role (the AC is not a sensor).

### hOn driver (write path)

- Load the device **command catalog** (`loadCommands`) and model options
  (`loadApplianceModel`) once per driver (cache with sensible invalidation), and translate
[...]

Issue #250, opened 2026-07-16
hOn AC read path: sensor readings + inspection


AGRFesta:
## Goal

The AC driver exposes the current device state through the existing outbound ports: ambient
readings as a `Sensor`, raw state as `Inspectable` for diagnostics.

## Scope

- **`HonAirConditioner` driver** (replacing the #249 stub) implementing:
  - **`Sensor.fetchReadings()`**: polls `GET /commands/v1/context` (`loadAttributes`) and
    maps the AC attributes to a `SensorReadings` implementation. The AC reports indoor
    temperature (`tempIndoor`) and — model permitting — humidity; map onto
    `ThermoHygroDataValue` so the existing sensor pipeline (assignments, history, home
    state) works unchanged. Verify the exact attribute keys against the real context
    payload captured in Phase 0 (addhOn's `sensor.py` / `const.py` are the key reference).
  - **`Inspectable.inspect()`**: returns the raw context payload (as `NetatmoSmarther`
    does with the room status) to ease debugging and future attribute mapping.
- **Attribute parsing is defensive**: hOn returns most values as strings inside
  `shadow.parameters` / `lastConnEvent` style structures whose shape can drift; absent or
  malformed values produce a typed failure or an empty reading, never an exception escaping
  the adapter. Shape confirmed live during Phase 0:
  `shadow.parameters.<key> = { "parNewVal": "<string value>", "lastUpdate": "<iso ts>" }`
  (top-level context keys: `resultCode`, `shadow`, `activity`, `commandHistory`,
  `lastConnEvent`).
- Keep the driver `runBlocking`-based like the other providers' drivers.

## Testing

Slice/unit tests with fixture payloads (real anonymized context captured in Phase 0):
happy-path mapping to `ThermoHygroData`, missing temperature attribute, malformed payload,
device offline shape. Guard test on the most important mapping (temperature value and
unit/scale) per project convention.

## Out of scope

- Writing state (#251 ), API endpoints (#252 ).
- Non-climate AC telemetry (energy, compressor frequency) — possible follow-up.

## DoD

`./gradlew build` green (incl. detekt). CHANGELOG (`Added`): hOn AC exposed as
temperature/humidity sensor. (#250)

AGRFesta:
Closing: invalidated by live device findings.

  Phase 0 assumed the AC exposes ambient readings mappable to ThermoHygroDataValue. Live context captures of a real AS35PBPHRA-PRE (both off and on +
[...]

PR #257, opened 2026-07-18
issue #249: hOn device discovery — provider, factory, AC prototypes w…

### LINKED ISSUE #249: hOn device discovery: provider, factory, catalog prototype
## Goal

hOn appliances show up in the device sync: `Provider.HON` exists in the domain, the
appliance list is mapped to `ProviderDeviceData`, and the AC model is known to the
`DeviceModelCatalog` so it gets its roles resolved.

## Scope

- **`Provider.HON`** added to the core enum (check for exhaustive `when`s and persistence
  enum mappings that need the new value).
- **`HonService : DevicesProvider`** (`@ConditionalOnHon`): `getAllDevices()` calls
  `loadAppliances()` (`POST /unified-api/v1/view/appliance-list` — the legacy
  `GET /commands/v1/appliance` returns `[]` for every account, do not use it) and maps each
  entry to `ProviderDeviceData`:
  - `deviceProviderId` = `macAddress`
  - `name` = the appliance nickname field
  - `model` = a stable model identifier from the list entry (decide between `modelName` /
    `applianceModelId` when inspecting the real payload captured in Phase 0; it must match
    what the prototype declares)
  - non-AC appliances are listed too (they will simply have no prototype/roles yet).
- **`HonDevicesFactory : ProviderDevicesFactory`**: creates the AC driver for the AC model
  (driver implemented in #250  — this issue can introduce it as a stub honouring
  `DeviceDriver`), unknown models get a no-op driver or are handled per the factory
  conventions of the other providers.
- **`DevicePrototype` bean** for the real AC model (values captured in Phase 0 of the epic),
  registered unconditionally (prototypes stay alive even when the provider is disabled),
  with `roles = {SENSOR, ACTUATOR}`.
- Note: the appliance-list entry also carries the fields needed later by the command
  endpoints (`applianceModelId`, `code`/`serialNumber` fallback rule, `eepromId`,
  `fwVersion`, `series`). Preserve them in the provider-side representation the factory
  hands to drivers, so #250/#251 don't need a second list call.

## Testing

Slice test: stubbed appliance-list response (use the real anonymized payload captured in
Phase 0 as fixture) → `getAllDevices()` returns the mapped `ProviderDeviceData`; defensive
behaviour on missing/malformed `modules.applianceList.payload.appliances` path. Catalog
test: the AC model resolves to the prototype with both roles.

## Out of scope

- Actual readings/commands (#250, #251).
- Prototypes for non-AC appliance types.

## DoD

`./gradlew build` green (incl. detekt). CHANGELOG (`Added`): hOn provider — appliances
discovered by device sync. (#249)

AGRFesta:
…ith appliance-ref store

### DIFF
--- a/providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonApplianceStore.kt
+++ b/providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonApplianceStore.kt
@@ -1,0 +1,21 @@
+package org.agrfesta.sh.api.providers.hon
+
+import org.springframework.stereotype.Service
+import java.util.concurrent.ConcurrentHashMap
+
+/**
+ * In-memory registry of the appliance-list fields the hOn command endpoints need
+ * ([HonApplianceRef]), keyed by mac address. [HonService] refreshes it on every device
+ * sync so factory-created drivers can reach those fields without a second list call.
+ */
+@Service
+@ConditionalOnHon
+class HonApplianceStore {
+    private val refs = ConcurrentHashMap<String, HonApplianceRef>()
+
+    fun save(ref: HonApplianceRef) {
+        refs[ref.macAddress] = ref
+    }
+
+    fun refOf(macAddress: String): HonApplianceRef? = refs[macAddress]
+}
--- a/providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonService.kt
+++ b/providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonService.kt
@@ -1,0 +1,74 @@
+package org.agrfesta.sh.api.providers.hon
+
+import arrow.core.Either
+import com.fasterxml.jackson.databind.JsonNode
+import kotlinx.coroutines.runBlocking
+import org.agrfesta.sh.api.core.application.ports.outbounds.devices.DevicesProvider
+import org.agrfesta.sh.api.core.domain.devices.DeviceModel
+import org.agrfesta.sh.api.core.domain.devices.Provider
+import org.agrfesta.sh.api.core.domain.devices.ProviderDeviceData
+import org.agrfesta.sh.api.core.domain.failures.DevicesProviderError
+import org.agrfesta.sh.api.core.domain.failures.DevicesProviderFailure
+import org.springframework.stereotype.Service
+
+@Service
+@ConditionalOnHon
+class HonService(
+    private val client: HonApiClient,
+    private val store: HonApplianceStore
+) : DevicesProvider {
+    override val provider: Provider = Provider.HON
+
+    override fun getAllDevices(): Either<DevicesProviderFailure, Collection<ProviderDeviceData>> = runBlocking {
+        client.loadAppliances()
+            .mapLeft { DevicesProviderError(it.toException()) }
+            .map { appliances -> appliances.mapNotNull { it.toProviderDeviceData() } }
+    }
+
+    /** Defensive: an entry without the identity fields cannot become a device — skip it. */
+    private fun JsonNode.toProviderDeviceData(): ProviderDeviceData? {
+        val macAddress = textOrNull("macAddress") ?: return null
+        val modelName = textOrNull("modelName") ?: return null
+        store.save(toApplianceRef(macAddress))
+        return ProviderDeviceData(
+            deviceProviderId = macAddress,
+            provider = provider,
+            name = textOrNull("nickName")?.trim() ?: modelName,
+            model = DeviceModel(MODEL_PREFIX + modelName),
+        )
+    }
+
+    /** The list-entry fields the command endpoints need (#250/#251), kept for the factory. */
+    private fun JsonNode.toApplianceRef(macAddress: String) = HonApplianceRef(
+        macAddress = macAddress,
+        applianceType = textOrNull("applianceTypeName").orEmpty(),
+        applianceModelId = textOrNull("applianceModelId").orEmpty(),
+        code = textOrNull("code") ?: codeFromSerialNumber(),
+        firmwareId = textOrNull("eepromId"),
+        fwVersion = textOrNull("fwVersion"),
+        series = textOrNull("series"),
+    )
+
+    /** addhOn fallback: the code is the serial's first 8 chars, 11 when the serial is 18+ long. */
+    private fun JsonNode.codeFromSerialNumber(): String {
+        val serialNumber = textOrNull("serialNumber").orEmpty()
+        val length = if (serialNumber.length < LONG_SERIAL_LENGTH) SHORT_SERIAL_CODE_LENGTH else SERIAL_CODE_LENGTH
+        return serialNumber.take(length)
+    }
+
+    private fun JsonNode.textOrNull(field: String): String? =
+        get(field)?.takeUnless { it.isNull }?.asText()?.takeIf { it.isNotBlank() }
+
+    companion object {
+        /** Provider qualifier of every hOn model string (e.g. `hon/AS25PBPHRA-PRE`). */
+        const val MODEL_PREFIX = "hon/"
+
+        /** Provider-qualified models of the known hOn AC appliances (captured in Phase 0). */
+        const val AC_AS25PBPHRA_PRE_MODEL = MODEL_PREFIX + "AS25PBPHRA-PRE"
+        const val AC_AS35PBPHRA_PRE_MODEL = MODEL_PREFIX + "AS35PBPHRA-PRE"
+
+        private const val LONG_SERIAL_LENGTH = 18
+        private const val SHORT_SERIAL_CODE_LENGTH = 8
+        private const val SERIAL_CODE_LENGTH = 11
+    }
+}
--- a/providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/devices/HonPrototypes.kt
+++ b/providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/devices/HonPrototypes.kt
@@ -1,0 +1,37 @@
+package org.agrfesta.sh.api.providers.hon.devices
+
+import org.agrfesta.sh.api.core.application.ports.outbounds.devices.DevicePrototype
+import org.agrfesta.sh.api.core.domain.devices.DeviceFeature
+import org.agrfesta.sh.api.core.domain.devices.DeviceModel
+import org.agrfesta.sh.api.core.domain.devices.Provider
+import org.agrfesta.sh.api.providers.hon.HonService
+import org.springframework.stereotype.Component
+
+/**
+ * hOn device prototypes. Registered as **unconditional** `@Component`s (not
+ * `@ConditionalOnHon`): the jar is always on the classpath, only credentials are conditional,
+ * so the read path stays alive when the provider is disabled. The model strings reuse
+ * [HonService]'s constants as the single source of truth, avoiding drift with the provider
+ * mapping.
+ *
+ * Roles are empty until [HonAc] implements the capability interfaces (SENSOR arrives with the
+ * readings issue, ACTUATOR with the commands issue — #250/#251): roles are user-visible
+ * behavior (feature filters, area assignments) and must match what the driver can actually do,
+ * as pinned by `DevicePrototypesContractTest` (precedent: `SwitchBotHubPrototype`).
+ */
+
+@Component
+class HonAcAs25Prototype : DevicePrototype {
+    override val model = DeviceModel(HonService.AC_AS25PBPHRA_PRE_MODEL)
+    override val provider = Provider.HON
+    override val driverType = HonAc::class
+    override val roles = emptySet<DeviceFeature>()
+}
+
+@Component
+class HonAcAs35Prototype : DevicePrototype {
+    override val model = DeviceModel(HonService.AC_AS35PBPHRA_PRE_MODEL)
+    override val provider = Provider.HON
+    override val driverType = HonAc::class
+    override val roles = emptySet<DeviceFeature>()
+}
--- a/providers/src/test/kotlin/org/agrfesta/sh/api/providers/hon/devices/HonDevicesFactoryTest.kt
+++ b/providers/src/test/kotlin/org/agrfesta/sh/api/providers/hon/devices/HonDevicesFactoryTest.kt
@@ -1,0 +1,52 @@
+package org.agrfesta.sh.api.providers.hon.devices
+
+import io.kotest.assertions.withClue
+import io.kotest.matchers.shouldBe
+import io.kotest.matchers.types.shouldBeInstanceOf
+import org.agrfesta.sh.api.core.domain.devices.DeviceModel
+import org.agrfesta.sh.api.core.domain.devices.Provider
+import org.agrfesta.sh.api.domain.aDevice
+import org.agrfesta.sh.api.providers.hon.HonApplianceStore
+import org.agrfesta.sh.api.providers.hon.HonService
+import org.junit.jupiter.api.Test
+
+class HonDevicesFactoryTest {
+    private val store = HonApplianceStore()
+    private val factory = HonDevicesFactory(store)
+
+    @Test
+    fun `builds the HonAc driver for the known AC models`() {
+        // Given
+        val as25 = aDevice(provider = Provider.HON, model = DeviceModel(HonService.AC_AS25PBPHRA_PRE_MODEL))
+        val as35 = aDevice(provider = Provider.HON, model = DeviceModel(HonService.AC_AS35PBPHRA_PRE_MODEL))
+
+        // When
+        val driver = factory.createDevice(as25)
+
+        // Then
+        withClue("the driver should carry the record's identity") {
+            driver.shouldBeInstanceOf<HonAc>()
+            driver.uuid shouldBe as25.uuid
+            driver.deviceProviderId shouldBe as25.deviceProviderId
+        }
+        withClue("both known AC models should get the AC driver") {
+            factory.createDevice(as35).shouldBeInstanceOf<HonAc>()
+        }
+    }
+
+    @Test
+    fun `builds a no-op driver for unknown hOn models`() {
+        // Given: an appliance type the provider lists but has no prototype for yet
+        val fridge = aDevice(provider = Provider.HON, model = DeviceModel("hon/REFRIGERATOR-X"))
+
+        // When
+        val driver = factory.createDevice(fridge)
+
+        // Then
+        withClue("an unknown model should get the no-op driver, still carrying the identity") {
+            driver.shouldBeInstanceOf<HonUnknownDevice>()
+            driver.uuid shouldBe fridge.uuid
+            driver.deviceProviderId shouldBe fridge.deviceProviderId
+        }
+    }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Add hOn provider to the device sync: appliances are discovered as devices; the known AC models are registered in the catalog (roles arrive with the readings/commands issues). (#249)
+
 ## [1.4.4] - 2026-07-18
 
 ### Added
--- a/providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonFailure.kt
+++ b/providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonFailure.kt
@@ -27,3 +27,19 @@ data class HonNetworkError(val reason: String?) : HonFailure
 
 /** The cloud answered with a body that is not valid JSON. */
 data object HonNonJsonResponse : HonFailure
+
+/**
+ * Maps each failure to an exception with a human-readable message, for the seams that surface
+ * `exception.message` to the user (diagnostics, sync reports). Mirrors Netatmo's
+ * `NetatmoClientFailure.toException()`.
+ */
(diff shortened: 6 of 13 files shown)


providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonService.kt
```diff
@@ -0,0 +1,74 @@
...
+
+    /** Defensive: an entry without the identity fields cannot become a device — skip it. */
+    private fun JsonNode.toProviderDeviceData(): ProviderDeviceData? {
+        val macAddress = textOrNull("macAddress") ?: return null
+        val modelName = textOrNull("modelName") ?: return null
+        store.save(toApplianceRef(macAddress))
+        return ProviderDeviceData(
+            deviceProviderId = macAddress,
+            provider = provider,
+            name = get("nickName").asText().trim().ifBlank { modelName },
```
AGRFesta: 🤖 **Reviewer** — _severity: high · confidence: high_

❗️ **Reachable NPE that escapes the adapter's exception boundary.** `get("nickName")` returns `null` when the field is absent, so `.asText()` throws. Unlike `SwitchBotService.getAllDevices` (which wraps the whole mapping in try/catch) there is no net here, so one appliance without a `nickName` key makes `getAllDevices` **throw** instead of returning `Left` — violating ARCHITECTURE.md's exception-boundary rule (adapters must catch technical exceptions and map them to domain failures) and breaking the `DevicesProvider` contract the sync relies on.

Two related edges, both fixed by routing through the helper you already have:
- absent `nickName` → NPE (above);
- `"nickName": null` → Jackson `NullNode.asText()` returns the string `"null"` → a device named `null`.

Suggestion: `name = textOrNull("nickName")?.trim()?.ifEmpty { null } ?: modelName` (or fold the trim into `textOrNull`). The KDoc on `toProviderDeviceData` promises defensiveness for identity fields — `nickName` deserves the same treatment, plus a test for the missing-`nickName` case (the current suite covers only the blank one).

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/devices/HonPrototypes.kt
```diff
@@ -0,0 +1,33 @@
...
+ * [HonService]'s constants as the single source of truth, avoiding drift with the provider
+ * mapping.
+ */
+
+@Component
+class HonAcAs25Prototype : DevicePrototype {
+    override val model = DeviceModel(HonService.AC_AS25PBPHRA_PRE_MODEL)
+    override val provider = Provider.HON
+    override val driverType = HonAc::class
+    override val roles = setOf(SENSOR, ACTUATOR)
```
AGRFesta: 🤖 **Reviewer** — _severity: high · confidence: high_

❗️ **The prototypes promise roles the driver cannot play — the exact disagreement `DevicePrototypesContractTest` exists to prevent.** `roles = setOf(SENSOR, ACTUATOR)`, but `HonAc` implements neither the `Sensor` nor the `Actuator` capability interface (readings/commands arrive with #250/#251, per its own KDoc). The shared contract test pins "each prototype's roles match the capability interfaces its driver implements" — these prototypes were not registered there, and **would fail it today**.

Because the prototypes are unconditional `@Component`s, the over-declared roles are live user-visible behavior right now, even with hOn disabled:
- `GET /devices?feature=SENSOR` (and `ACTUATOR`) lists the ACs;
- `AssignSensorToAreaService` / `AssignActuatorToAreaService` accept assigning an AC to an area;
- …yet `FetchSensorReadingsService` silently skips the driver (`driver !is Sensor`) and heating resolution drops it (`as? Heater`) — dead assignments with silent no-ops, and no signal to the user about why the AC never produces readings.

Recommendation: declare `roles = emptySet()` now (precedent: `SwitchBotHubPrototype`) and expand to `SENSOR`/`ACTUATOR` in the same PRs that make `HonAc` actually implement the capabilities. That also lets these prototypes join the contract test immediately (see companion comment on the factory test).

providers/src/test/kotlin/org/agrfesta/sh/api/providers/hon/devices/HonDevicesFactoryTest.kt
```diff
@@ -0,0 +1,52 @@
...
+import io.kotest.matchers.types.shouldBeInstanceOf
+import org.agrfesta.sh.api.core.domain.devices.DeviceModel
+import org.agrfesta.sh.api.core.domain.devices.Provider
+import org.agrfesta.sh.api.domain.aDevice
+import org.agrfesta.sh.api.providers.hon.HonApplianceStore
+import org.agrfesta.sh.api.providers.hon.HonService
+import org.junit.jupiter.api.Test
+
+class HonDevicesFactoryTest {
+    private val store = HonApplianceStore()
```
AGRFesta, line 14: 🤖 **Reviewer** — _severity: medium · confidence: high_

**Register the hOn factory and prototypes in `DevicePrototypesContractTest`.** That test declares itself the single pinning point for the `model → driver` and `roles → capabilities` couplings (its KDoc: "a prototype can never silently disagree with the factory it describes, nor with the capabilities its driver actually implements"). This PR adds a new factory and two prototypes but leaves them out of both the `factories` map and the `prototypes` list there. The bespoke tests here re-cover part of contract 1 (driver type), but nothing enforces contract 2 (roles ↔ capabilities) for hOn — which is currently violated (see comment on `HonPrototypes.kt`). Adding them would have caught that at build time, which is exactly the deterministic-enforcement layer this project prefers over review-time catches.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonService.kt
```diff
@@ -0,0 +1,74 @@
...
+@ConditionalOnHon
+class HonService(
+    private val client: HonApiClient,
+    private val store: HonApplianceStore
+) : DevicesProvider {
+    override val provider: Provider = Provider.HON
+
+    override fun getAllDevices(): Either<DevicesProviderFailure, Collection<ProviderDeviceData>> = runBlocking {
+        client.loadAppliances()
+            .mapLeft { DevicesProviderError(IllegalStateException(it.toString())) }
```
AGRFesta: 🤖 **Reviewer** — _severity: low · confidence: medium_

**Failure mapping stringifies the typed failure.** `DevicesProviderError(IllegalStateException(it.toString()))` throws away the `HonFailure` structure. Netatmo established the pattern for this exact seam: `NetatmoClientFailure.toException()` maps each variant to an exception with a readable message. This matters downstream: `InspectDeviceService` (and diagnostics generally) surfaces `failure.exception.message`, which for hOn will be a raw data-class `toString` (e.g. `HonLoginFlowBroken(reason=…)`). A small `HonFailure.toException()` extension would give parity with the sibling provider and human-readable diagnostics for the auth failures you already model (`HonMfaRequired`, `HonLoginFlowBroken`).

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonApplianceStore.kt
```diff
@@ -0,0 +1,21 @@
...
+@Service
+@ConditionalOnHon
+class HonApplianceStore {
+    private val refs = ConcurrentHashMap<String, HonApplianceRef>()
+
+    fun save(ref: HonApplianceRef) {
+        refs[ref.macAddress] = ref
+    }
+
+    fun refOf(macAddress: String): HonApplianceRef? = refs[macAddress]
```
AGRFesta, line 20: 🤖 **Reviewer** — _severity: low · confidence: medium_

**Flag for #250/#251: define the miss behavior of `refOf` before drivers depend on it.** The store is populated only by a successful sync in the same JVM, so after a restart every factory-built `HonAc` sees `refOf == null` until the first appliance-list call completes; entries for appliances removed from the account also linger forever (no eviction). Both are harmless in this PR — `HonAc` doesn't read the store yet — but the readings/commands work will hit the empty-store window on day one (e.g. a diagnostics or readings call scheduled before the first sync). Worth deciding now whether a miss triggers a lazy refresh or a typed failure, so #250 doesn't inherit an implicit ordering dependency between the sync scheduler and every other hOn code path.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonService.kt
```diff
@@ -0,0 +1,74 @@
...
+            model = DeviceModel(MODEL_PREFIX + modelName),
+        )
+    }
+
+    /** The list-entry fields the command endpoints need (#250/#251), kept for the factory. */
+    private fun JsonNode.toApplianceRef(macAddress: String) = HonApplianceRef(
+        macAddress = macAddress,
+        applianceType = textOrNull("applianceTypeName").orEmpty(),
+        applianceModelId = textOrNull("applianceModelId").orEmpty(),
+        code = textOrNull("code") ?: codeFromSerialNumber(),
```
AGRFesta, line 46: 🤖 **Reviewer** — _severity: low · confidence: low_

**Silent empty `code` when both `code` and `serialNumber` are missing.** `codeFromSerialNumber()` on an absent serial returns `""`, so the ref is stored with an empty code and the command endpoints (#250/#251) will eventually send it as-is. Since `HonApplianceRef.code` already defaults to `""` this may be intentional, but consider treating it as absent (keep the field empty only via the default) or logging at store time, so a malformed cloud entry is diagnosable at sync rather than as a mysterious command failure later.

AGRFesta:
🤖 **Reviewer** — review summary (round 1, sha `ba8d69191122b91629b5b4b7764391eddee564fb`)

Round 1 review of the hOn device discovery (provider + factory + AC prototypes + appliance-ref store). Solid overall: the provider mirrors the Netatmo/SwitchBot shape, the appliance-ref store is a sensible answer to the second-list-call problem, tests are behavior-driven with the real captured fixture, and CHANGELOG is in order (no endpoint changes, so no API-doc work due). Two significant findings: (1) a reachable NPE in the nickName mapping that escapes the adapter's exception boundary, and (2) the AC prototypes declare SENSOR+ACTUATOR roles that HonAc does not implement yet — precisely the disagreement DevicePrototypesContractTest exists to pin, and the new prototypes/factory were not registered there. Remaining findings are low-severity hardening/consistency notes. Side note, no action needed here: HonServiceTest imports BehaviorRegistry/ResponseSpec from the netatmo package — a third consumer now, reinforcing the already-tracked follow-up from PR #256 to hoist them to the providers test root.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonService.kt
AGRFesta: 🧑‍💻 **Author** — Accepted, fixed in 49e47e8. `name` now routes through the safe helper (`textOrNull("nickName")?.trim() ?: modelName`), and `textOrNull` gained a `takeUnless { it.isNull }` guard, so both edges are covered: a **missing** `nickName` falls back to the model name instead of throwing, and an **explicit `"nickName": null`** no longer becomes a device named `"null"`. Two RED-first tests pin the behaviors (missing and explicit-null nickname); the `NullNode` guard also protects the identity fields (`macAddress`/`modelName`) from explicit nulls.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/devices/HonPrototypes.kt
AGRFesta: 🧑‍💻 **Author** — Accepted, fixed in 49e47e8. Both prototypes now declare `roles = emptySet()` (precedent: `SwitchBotHubPrototype`), with a KDoc note that SENSOR arrives with the readings issue (#250) and ACTUATOR with the commands issue (#251), alongside the real capability implementations on `HonAc`. `HonPrototypesTest` pins the empty roles, the CHANGELOG line was realigned, and — per the companion comment — the hOn factory and prototypes are now registered in `DevicePrototypesContractTest`, so expanding roles without the capabilities fails the build. Worth noting the roles were over-declared because the issue text asked for `{SENSOR, ACTUATOR}` upfront; the contract test is the better source of truth here.

providers/src/test/kotlin/org/agrfesta/sh/api/providers/hon/devices/HonDevicesFactoryTest.kt
AGRFesta: 🧑‍💻 **Author** — Accepted, fixed in 49e47e8. `HonDevicesFactory(HonApplianceStore())` joined the `factories` map and both AC prototypes the `prototypes` list. With the roles emptied (see the `HonPrototypes.kt` thread) both contracts pass; from now on expanding the roles in #250/#251 without implementing the capability interfaces on `HonAc` fails the build — exactly the deterministic enforcement this test exists for.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonService.kt
AGRFesta: 🧑‍💻 **Author** — Accepted, fixed in 49e47e8. Added `internal fun HonFailure.toException()` mirroring `NetatmoClientFailure.toException()`: each variant maps to a readable message (e.g. `hOn server error (HTTP 500)`, `hOn account requires email-OTP MFA: cannot authenticate unattended`, `hOn login flow broken: <reason>`), and `HonService` maps failures through it. The existing failure test now asserts on the surfaced message rather than just the failure type.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonApplianceStore.kt
AGRFesta: 🧑‍💻 **Author** — Agreed, and deciding it now so #250 doesn't inherit an implicit ordering dependency: (1) on a `refOf` miss the AC driver (#250) will fall back to a lazy `loadAppliances()` refresh through `HonApiClient` instead of failing — the store is a cache of the list call, not a source of truth — with a typed failure only if the appliance is absent from the refreshed list too; (2) eviction lands with the same issue: the sync will rebuild the store content instead of only upserting, so appliances removed from the account disappear. Tracked in the #250 notes; no behavior change in this PR since `HonAc` does not read the store yet.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonService.kt
AGRFesta: 🧑‍💻 **Author** — Keeping the empty-string default for now: the ref is inert in this PR (nothing consumes `code` yet), and the right diagnosis point is where the field is used — #251's command path will validate the ref before calling the cloud and surface a typed failure naming the missing field, which is more actionable than a sync-time log line (the provider adapter has no logging infrastructure today). The addhOn-derived fallback already covers every shape observed in the real payloads: both fixture appliances carry `code` and `serialNumber`.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonService.kt
```diff
@@ -0,0 +1,74 @@
...
+    private fun JsonNode.toProviderDeviceData(): ProviderDeviceData? {
+        val macAddress = textOrNull("macAddress") ?: return null
+        val modelName = textOrNull("modelName") ?: return null
+        store.save(toApplianceRef(macAddress))
+        return ProviderDeviceData(
+            deviceProviderId = macAddress,
+            provider = provider,
+            name = get("nickName").asText().trim().ifBlank { modelName },
+            model = DeviceModel(MODEL_PREFIX + modelName),
+        )
```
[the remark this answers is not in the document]
AGRFesta: 🧑‍💻 **Author** — Fixed in 49e47e8: `name` now routes through the safe `textOrNull` helper with a `modelName` fallback, covered by RED-first tests for both the missing and the explicit-null `nickName`. Thanks!

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonService.kt
```diff
@@ -0,0 +1,74 @@
...
+
+    /** addhOn fallback: the code is the serial's first 8 chars, 11 when the serial is 18+ long. */
+    private fun JsonNode.codeFromSerialNumber(): String {
+        val serialNumber = textOrNull("serialNumber").orEmpty()
+        val length = if (serialNumber.length < LONG_SERIAL_LENGTH) SHORT_SERIAL_CODE_LENGTH else SERIAL_CODE_LENGTH
+        return serialNumber.take(length)
+    }
+
+    private fun JsonNode.textOrNull(field: String): String? =
+        get(field)?.asText()?.takeIf { it.isNotBlank() }
```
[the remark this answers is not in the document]
AGRFesta: 🧑‍💻 **Author** — Fixed in 49e47e8: `textOrNull` now filters `NullNode` via `takeUnless { it.isNull }` exactly as suggested, pinned by a test where `"nickName": null` falls back to the model name instead of becoming the string `"null"`. Thanks!

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonService.kt
AGRFesta: 🤖 **Reviewer**

Verified in 49e47e8: `name` routes through `textOrNull`, which now filters `NullNode` before `asText()`, so both the missing-key NPE and the explicit-null-to-"null" edge are closed, each pinned by its own test. The helper's ordering also preserves the round-1 blank-nickname fallback (`isNotBlank` filters whitespace-only values before the `trim`), and the guard extends to the identity fields for free. Exception boundary restored.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/devices/HonPrototypes.kt
AGRFesta: 🤖 **Reviewer**

Verified in 49e47e8: both prototypes declare `roles = emptySet()`, the KDoc documents the deferral to #250/#251 with the contract-test rationale, `HonPrototypesTest` pins the empty roles, and the CHANGELOG line was realigned to the actual behavior. With the contract-test registration (companion thread) the expansion path is now build-enforced. Agreed that the issue text's upfront `{SENSOR, ACTUATOR}` was the source of the over-declaration — the contract test is the better source of truth.

providers/src/test/kotlin/org/agrfesta/sh/api/providers/hon/devices/HonDevicesFactoryTest.kt
AGRFesta: 🤖 **Reviewer**

Verified in 49e47e8: `HonDevicesFactory(HonApplianceStore())` is in the `factories` map and both AC prototypes are in the `prototypes` list; `DevicePrototypesContractTest` passes. Expanding roles in #250/#251 without implementing the capability interfaces on `HonAc` now fails the build — the deterministic layer is in place.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonService.kt
AGRFesta: 🤖 **Reviewer**

Verified in 49e47e8: `HonFailure.toException()` is exhaustive over all eight variants with human-readable messages and mirrors the Netatmo precedent faithfully (same file placement, `internal`, per-variant messages), and the failure test now asserts the surfaced message (`hOn server error (HTTP 500)`) rather than just the type. Diagnostics parity achieved.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonApplianceStore.kt
AGRFesta: 🤖 **Reviewer**

Resolved: the decision is exactly what this thread asked for — treating the store as a cache with a lazy `loadAppliances()` refresh on miss (typed failure only if the appliance is absent from the refreshed list) removes the ordering dependency, and rebuild-instead-of-upsert in #250 covers eviction. No behavior change needed in this PR; tracked in the #250 notes.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/HonService.kt
AGRFesta: 🤖 **Reviewer**

Conceded. The ref is inert in this PR, and validating at the #251 use point with a typed failure naming the missing field is more actionable than a sync-time log line the adapter has no infrastructure for. The fixture evidence (both real appliances carry `code` and `serialNumber`) supports deferring. Holding you to the #251 validation, though — that's now the load-bearing half of this rejection.

AGRFesta:
🤖 **Reviewer** — review summary (round 2, sha `49e47e8c5dcae9d7f0f4392fa7a2a3aa30caee0e`)

Round 2 re-review, scoped to the six open reviewer threads plus the delta (ba8d691..49e47e8). All four accepted findings are verified fixed in 49e47e8: the nickName NPE and NullNode edge are closed through the hardened textOrNull helper with RED-first tests for both edges; the prototypes now declare empty roles with the deferral documented in KDoc and pinned by HonPrototypesTest; the hOn factory and both AC prototypes are registered in DevicePrototypesContractTest (verified passing), making roles-vs-capabilities build-enforced for hOn; and HonFailure.toException() mirrors the Netatmo precedent exhaustively with the surfaced message asserted in test. The appliance-store follow-up decision (lazy refresh on miss, rebuild-based eviction in #250) resolves that thread, and the rejection on the empty code field is conceded: validating at the #251 use point with a typed failure is the more actionable seam. No new findings in the delta. All threads resolved — nothing outstanding from the reviewer side.

Issue #252, opened 2026-07-16
AC control REST endpoints + API docs + Bruno


AGRFesta:
## Goal

Expose AC control through the API so the dashboard/clients can drive the air conditioner:
read its control state and change power, mode, target temperature and fan speed.

## Scope

- **Inbound use cases** (one service class per use case interface, `execute()` naming, per
  project convention) in the appropriate subdomain package:
  - get AC control state (power, mode, target temperature, fan speed — from the driver);
  - set AC settings (partial update: only the provided fields change, the driver's
    read-modify-write fills the rest).
- **Endpoints** (proposal — final shape settled at Phase 0):
  - `GET /devices/{uuid}/air-conditioner` → current control state;
  - `PATCH /devices/{uuid}/air-conditioner` with body
    `{ "power": "ON", "mode": "COOL", "targetTemperature": 22, "fanSpeed": "AUTO" }`
    (all fields optional, at least one required).
  - Validation errors (unknown mode, temperature out of the device's catalog range) map to
    400 with a meaningful message; driver/cloud failures map per the existing failure→HTTP
    conventions; non-AC device uuid → 404/409 per existing conventions.
- **Docs & collection** (project convention):
  - `docs/api/air-conditioner.md` (or extend the devices doc — follow the existing
    resource split), row in `docs/api/API_INDEX.md`;
  - Bruno collection `bruno/air-conditioner/` (or under `bruno/devices/`), one `.bru` per
    endpoint.
- **CHANGELOG** (`Added`): AC control endpoints — power, mode, target temperature, fan
  speed for hOn air conditioners. (#252)

## Testing

Controller slice tests (`@WebMvcTest`): happy paths, partial update, validation failures,
failure mapping, device-not-an-AC case. Use-case unit tests with a mocked `AirConditioner`
port.

## Out of scope

- Any automation using the AC (climate strategies, schedules) — future work on top of the
  port.
- UI work.

## DoD

`./gradlew build` green (incl. detekt). API docs + Bruno + CHANGELOG updated in this issue.

PR #259, opened 2026-07-19
issue #251: hOn AC write path — AirConditioner port, catalog-driven s…

### LINKED ISSUE #251: AC control: `AirConditioner` outbound port + hOn write path

AGRFesta:
…ettings RMW, raw-state inspection

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Expose the hOn AC raw cloud state through `GET /devices/{uuid}/diagnostics`: the AC driver is now inspectable per device, like the SwitchBot and Netatmo drivers. (#251)
+- hOn AC devices now expose the `ACTUATOR` feature in device responses. (#251)
+
 ## [1.4.5] - 2026-07-18
 
 ### Added
--- a/core/src/main/kotlin/org/agrfesta/sh/api/core/application/ports/outbounds/devices/AirConditioner.kt
+++ b/core/src/main/kotlin/org/agrfesta/sh/api/core/application/ports/outbounds/devices/AirConditioner.kt
@@ -1,0 +1,19 @@
+package org.agrfesta.sh.api.core.application.ports.outbounds.devices
+
+import arrow.core.Either
+import org.agrfesta.sh.api.core.domain.commons.Temperature
+import org.agrfesta.sh.api.core.domain.devices.AcFanSpeed
+import org.agrfesta.sh.api.core.domain.devices.AcMode
+import org.agrfesta.sh.api.core.domain.failures.ActuatorOperationFailure
+
+/**
+ * An air conditioner: an on/off actuator that can additionally be driven on operating mode,
+ * target temperature and fan speed. Each setter is a complete, self-contained command: the
+ * driver is responsible for composing it with the device's current state when the underlying
+ * protocol transmits full-state writes (as hOn's `settings` command does).
+ */
+interface AirConditioner : OnOffActuator {
+    fun setMode(mode: AcMode): Either<ActuatorOperationFailure, Unit>
+    fun setTargetTemperature(temperature: Temperature): Either<ActuatorOperationFailure, Unit>
+    fun setFanSpeed(speed: AcFanSpeed): Either<ActuatorOperationFailure, Unit>
+}
--- a/providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/devices/AcSettingsCommand.kt
+++ b/providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/devices/AcSettingsCommand.kt
@@ -1,0 +1,157 @@
+package org.agrfesta.sh.api.providers.hon.devices
+
+import arrow.core.Either
+import com.fasterxml.jackson.databind.JsonNode
+import java.math.BigDecimal
+
+/**
+ * The hOn AC `settings` command as a read-modify-write value object.
+ *
+ * hOn's `settings` command always transmits ALL of its parameters, so a write is never a
+ * single-field update: the full parameter set is rebuilt from the device command catalog
+ * (`GET /commands/v1/retrieve`, `settings.setParameters`) overlaid with the latest device
+ * state (`GET /commands/v1/context`), and the requested change is validated against the
+ * catalog rule of its parameter (enum membership, range/step). Ported from the addhOn
+ * engine semantics (`HonCommand` + `ac_command.py`).
+ */
+class AcSettingsCommand private constructor(
+    private val catalog: JsonNode,
+    private val context: JsonNode,
+    private val changes: Map<String, String>,
+) {
+    constructor(catalog: JsonNode, context: JsonNode) : this(catalog, context, emptyMap())
+
+    /**
+     * Returns a copy carrying the requested change, validated against the parameter's
+     * catalog rule (enum membership, range/step). The change wins over the current state.
+     */
+    fun with(parameter: String, value: String): Either<HonAcFailure, AcSettingsCommand> =
+        when (val failure = validationFailureOf(parameter, value)) {
+            null -> Either.Right(AcSettingsCommand(catalog, context, changes + (parameter to value)))
+            else -> Either.Left(failure)
+        }
+
+    /** The catalog rule the change breaks, or null when it is admitted. */
+    private fun validationFailureOf(parameter: String, value: String): HonAcFailure? {
+        val spec = catalog.at("/settings/setParameters/parameters/$parameter")
+        val allowed = spec.enumValues()
+        return when {
+            spec.isMissingNode -> AcUnknownParameter(parameter)
+            allowed != null && value !in allowed -> AcSettingNotSupported(parameter, value, allowed)
+            spec.get("typology")?.asText() == "range" && !spec.admitsInRange(value) ->
+                AcSettingOutOfRange(parameter, value)
+            else -> null
+        }
+    }
+
+    /**
+     * The admitted enum values as strings (the cloud sends them as JSON numbers), or null.
+     * An EMPTY `enumValues` is catalog drift, not a rule: null too — otherwise it would
+     * reject every value and crash the sanitization fallbacks.
+     */
+    private fun JsonNode.enumValues(): List<String>? =
+        get("enumValues")?.takeIf { it.isArray }?.map { it.asText() }?.takeIf { it.isNotEmpty() }
+
+    /**
+     * True when [value] falls inside this range spec's `minimumValue..maximumValue` AND lands
+     * on the `minimumValue + n * incrementValue` grid (exact [BigDecimal] arithmetic — no
+     * float epsilon needed). A missing/zero increment means "no declared grid" (addhOn quirk).
+     * Missing or non-numeric BOUNDS are catalog drift: the rule is unenforceable, so the
+     * value is admitted rather than bricking the control — the cloud stays the ultimate
+     * validator (a bad value comes back as a rejected command, still a typed failure).
+     */
+    private fun JsonNode.admitsInRange(value: String): Boolean {
+        val requested = value.toBigDecimalOrNull() ?: return false
+        val min = get("minimumValue")?.asText()?.toBigDecimalOrNull()
+        val max = get("maximumValue")?.asText()?.toBigDecimalOrNull()
+        if (min == null || max == null) return true
+        val step = get("incrementValue")?.asText()?.toBigDecimalOrNull()
+        val onGrid = step == null || step.signum() <= 0 ||
+            (requested - min).remainder(step).signum() == 0
+        return requested in min..max && onGrid
+    }
+
+    /** The full `parameters` block of the send body: one entry per catalog parameter. */
+    fun parameters(): Map<String, String> = buildMap {
+        catalog.at("/settings/setParameters/parameters").fields().forEach { (name, spec) ->
+            val base = sanitizedWind(name, contextValueOf(name, spec) ?: spec.defaultOf(), spec)
+            put(name, changes[name] ?: base)
+        }
+    }
+
+    /**
+     * Resets an off-enum wind direction to an admitted value: the device reports `"0"` while
+     * off, the cloud rejects the whole command on it (port of addhOn's
+     * `sanitize_wind_direction`). Vertical prefers the FIXED position `"2"` (never the swing
+     * `"8"`), horizontal the first non-`"0"`. Applied to the device-reported base only — a
+     * requested change wins untouched. No enum info in the catalog → nothing to sanitize
+     * against, value passes through.
+     */
+    private fun sanitizedWind(name: String, value: String, spec: JsonNode): String {
+        if (name !in WIND_DIRECTION_PARAMETERS) return value
+        val allowed = spec.enumValues() ?: return value
+        // Vertical prefers the fixed position 2 among the non-swing values; never 8 (swing).
+        val fixedVertical = (allowed - WIND_SWING_VERTICAL).let { fixed ->
+            when {
+                WIND_FIXED_VERTICAL in fixed -> WIND_FIXED_VERTICAL
+                else -> fixed.firstOrNull() ?: WIND_SWING_VERTICAL
+            }
+        }
+        return when {
+            value in allowed -> value
+            name == WIND_DIRECTION_VERTICAL -> fixedVertical
+            else -> allowed.firstOrNull { it != "0" } ?: allowed.first()
+        }
+    }
+
+    /**
+     * The `ancillaryParameters` block of the send body. `category == "rule"` entries
+     * (programRules) are engine metadata whose `fixedValue` is an object, not a wire value:
+     * addhOn strips them before sending, and so do we.
+     */
+    fun ancillaryParameters(): Map<String, String> = buildMap {
+        catalog.at("/settings/setParameters/ancillaryParameters").fields().forEach { (name, spec) ->
+            if (spec.get("category")?.asText() != "rule") {
+                put(name, contextValueOf(name, spec) ?: spec.defaultOf())
+            }
+        }
+    }
+
+    /** Latest device state of a parameter (`shadow.parameters.<name>.parNewVal`), if reported. */
+    private fun contextValueOf(name: String, spec: JsonNode): String? {
+        val reported = context.at("/shadow/parameters/$name/parNewVal")
+            .takeUnless { it.isMissingNode }?.asText() ?: return null
+        return if (spec.get("typology")?.asText() == "range") reported.normalizedDecimal() else reported
+    }
+
+    /**
+     * Strips insignificant decimals from a range value: the context reports `"26.00"` but the
+     * catalog grid is declared in plain form (`"16".."30"` step `"1"`) — mirror of addhOn's
+     * `str_to_float` normalization. Non-numeric values pass through untouched.
+     */
+    private fun String.normalizedDecimal(): String =
+        runCatching { BigDecimal(this).stripTrailingZeros().toPlainString() }.getOrDefault(this)
+
+    /**
+     * Catalog default of a parameter: `fixedValue` for fixed typology, `defaultValue`
+     * otherwise — falling back to `"0"` when either is absent (an enum may omit it, e.g. the
+     * real `windDirectionVerticalPositionSequence`), like addhOn's base parameter does.
+     */
+    private fun JsonNode.defaultOf(): String =
+        if (get("typology")?.asText() == "fixed") {
+            get("fixedValue")?.asText() ?: "0"
+        } else {
+            get("defaultValue")?.asText() ?: "0"
+        }
+
+    companion object {
+        private const val WIND_DIRECTION_VERTICAL = "windDirectionVertical"
+        private val WIND_DIRECTION_PARAMETERS = setOf(WIND_DIRECTION_VERTICAL, "windDirectionHorizontal")
+
+        /** Vertical position 8 = swing (oscillation): never a valid "fixed" fallback. */
+        private const val WIND_SWING_VERTICAL = "8"
+
+        /** The preferred fixed vertical position when resetting an off-enum value. */
+        private const val WIND_FIXED_VERTICAL = "2"
+    }
+}
--- a/providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/devices/HonAc.kt
+++ b/providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/devices/HonAc.kt
@@ -1,19 +1,118 @@
 package org.agrfesta.sh.api.providers.hon.devices
 
-import org.agrfesta.sh.api.core.application.ports.outbounds.devices.DeviceDriver
+import arrow.core.Either
+import arrow.core.raise.either
+import arrow.core.raise.ensureNotNull
+import kotlinx.coroutines.runBlocking
+import org.agrfesta.sh.api.core.application.ports.outbounds.devices.AirConditioner
+import org.agrfesta.sh.api.core.application.ports.outbounds.devices.Inspectable
+import org.agrfesta.sh.api.core.domain.commons.Temperature
+import org.agrfesta.sh.api.core.domain.devices.AcFanSpeed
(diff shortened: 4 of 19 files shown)


core/src/main/kotlin/org/agrfesta/sh/api/core/application/ports/outbounds/devices/AirConditioner.kt
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * An air conditioner: an on/off actuator that can additionally be driven on operating mode,
+ * target temperature and fan speed. Each setter is a complete, self-contained command: the
+ * driver is responsible for composing it with the device's current state when the underlying
+ * protocol transmits full-state writes (as hOn's `settings` command does).
+ */
+interface AirConditioner : OnOffActuator {
+    fun setMode(mode: AcMode): Either<ActuatorOperationFailure, Unit>
+    fun setTargetTemperature(temperature: BigDecimal): Either<ActuatorOperationFailure, Unit>
```
AGRFesta: 🤖 **Reviewer** — _severity: medium · confidence: high_

**Value Object rule — raw `BigDecimal` for a domain quantity.** `docs/ARCHITECTURE.md` (§Value Objects) mandates inline value classes for domain quantities, and `Temperature` already exists (BigDecimal-based, 2-decimal HALF_UP, used across the heating path). A target temperature on a core outbound port is exactly the quantity `Temperature` models.

Beyond consistency, the raw type leaks a real edge: a caller can pass an unscaled `BigDecimal` (e.g. `23.000000001`), which sails through the port and only dies deep in the driver as an opaque `AcSettingOutOfRange` from the catalog grid check — instead of being normalized at the domain boundary. Note the driver already compensates with `stripTrailingZeros().toPlainString()`, which is precisely `Temperature.of`'s normalization re-implemented.

Suggestion: `fun setTargetTemperature(temperature: Temperature)`, with the driver serializing `temperature.value`.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/devices/AcSettingsCommand.kt
```diff
@@ -0,0 +1,149 @@
...
+        get("enumValues")?.takeIf { it.isArray }?.map { it.asText() }
+
+    /**
+     * True when [value] falls inside this range spec's `minimumValue..maximumValue` AND lands
+     * on the `minimumValue + n * incrementValue` grid (exact [BigDecimal] arithmetic — no
+     * float epsilon needed). A missing/zero increment means "no declared grid" (addhOn quirk).
+     */
+    private fun JsonNode.admitsInRange(value: String): Boolean {
+        val requested = value.toBigDecimalOrNull() ?: return false
+        val min = get("minimumValue").asText().toBigDecimal()
```
AGRFesta: 🤖 **Reviewer** — _severity: medium · confidence: medium_

**Exception boundary — catalog-shape drift can throw through the outbound port.** Several reads here assume the cloud catalog is well-formed:

* `get("minimumValue").asText()` → NPE if a `range` spec omits `minimumValue` (same for `maximumValue`);
* `.toBigDecimal()` → `NumberFormatException` on a non-numeric bound;
* `defaultOf()` (line 134) → NPE on a `fixed` spec without `fixedValue`.

The catalog is external input, and this codebase itself documents the risk (`HonApiClient.loadAppliances`: "the cloud changes shapes without notice — fall back to empty"). Nothing between here and the core catches: Arrow's `either { }` in `HonAc.applySetting` does **not** catch exceptions, so a malformed spec would propagate a raw NPE out of the `AirConditioner` port into the application layer — violating the Exception Boundary rule in `docs/ARCHITECTURE.md` ("Must catch infrastructure/technical exceptions and map them to the expected Domain Errors"; "DON'T let exceptions propagate beyond the Adapter layer").

Suggestion: either make the individual reads defensive (missing/garbage bound → treat the spec as invalid → typed `HonAcFailure`), or wrap the command composition in `HonAc.applySetting` with a `runCatching`-style boundary mapping to a typed failure. A small unit test with a truncated range spec would pin it.

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Expose the hOn AC raw cloud state through `GET /devices/{uuid}/diagnostics`: the AC driver is now inspectable per device, like the SwitchBot and Netatmo drivers. (#251)
```
AGRFesta, line 11: 🤖 **Reviewer** — _severity: low · confidence: medium_

**Changelog misses the second observable behaviour of this PR.** The prototype roles change (`emptySet` → `ACTUATOR` in `HonPrototypes.kt`) is user-visible: per `docs/domain/DEVICES.md` the `features` array in API responses is a projection of the catalog roles, so hOn ACs now expose `ACTUATOR` in `GET /devices` and become eligible for actuator-scoped filters. That's an observable behaviour with no entry — the CLAUDE.md changelog convention asks for one line under `Added` (e.g. "hOn AC devices now expose the `ACTUATOR` feature. (#251)").

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/devices/HonAc.kt
```diff
@@ -1,19 +1,124 @@
...
+    private val client: HonApiClient,
+) : AirConditioner, Inspectable {
     override val provider: Provider = Provider.HON
+
+    /**
+     * The raw `GET /commands/v1/context` body, **verbatim** (pass-through like SwitchBot: no
+     * parse -> re-serialize round trip), so the diagnostics endpoint exposes exactly what the
+     * cloud sent.
+     */
+    override fun inspect(): Either<DevicesProviderFailure, String> =
```
AGRFesta, line 41: 🤖 **Reviewer** — _severity: low · confidence: medium_

**API doc not updated for the extended diagnostics behaviour.** `docs/api/devices.md` §`GET /devices/{uuid}/diagnostics` explains the provider-specific response shape by example — "SwitchBot returns the device status; Netatmo returns the device's room status". hOn is now a third inspectable provider (raw `/commands/v1/context` body, verbatim), and the changelog entry advertises exactly that. Per the CLAUDE.md API-docs convention (endpoint behaviour modified → update `docs/api/<resource>.md`), the example list should name hOn too. One sentence is enough.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/devices/HonAc.kt
```diff
@@ -1,19 +1,124 @@
...
+                client.sendCommand(
+                    appliance = ref,
+                    command = SETTINGS_COMMAND,
+                    parameters = command.parameters(),
+                    ancillaryParameters = command.ancillaryParameters(),
+                ).mapLeft(::HonAcProviderFailure).bind()
+            }
+        }
+
+    companion object {
```
AGRFesta, line 115: 🤖 **Reviewer** — _severity: low · confidence: high_

**Structure: companion object stranded mid-class.** Two interface overrides (`setTargetTemperature`, `setFanSpeed`) trail *after* the companion object — this reads like an un-tidied TDD append (the Phase 3 sweep should have caught it). Group the `AirConditioner` overrides together with `on`/`off`/`setMode` and move the companion to the bottom, per Kotlin conventions.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/devices/HonAc.kt
```diff
@@ -1,19 +1,124 @@
...
+     * cloud sent.
+     */
+    override fun inspect(): Either<DevicesProviderFailure, String> =
+        runBlocking {
+            either {
+                val ref = ensureNotNull(appliances.refOf(deviceProviderId)) {
+                    DevicesProviderError(IllegalStateException("Appliance '$deviceProviderId' not synced yet"))
+                }
+                client.rawContext(
+                    mapOf(
```
AGRFesta: 🤖 **Reviewer** — _severity: low · confidence: medium_

**Duplicated wire-query encoding.** This param map re-encodes exactly what `HonApiClient.loadAttributes` already builds (`macAddress`/`applianceType`/`category=CYCLE`). Two encodings of the same query will drift silently if the category or a param name ever changes — and only one of them is covered by the driver test. Consider a `rawContext(appliance: HonApplianceRef)` overload on the client so the context-query shape lives in one place (mirroring how `loadAttributes` owns it for the parsed variant).

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/devices/HonAc.kt
```diff
@@ -1,19 +1,124 @@
...
+        AcFanSpeed.LOW -> "3"
+        AcFanSpeed.AUTO -> "5"
+    }
+
+    /** One read-modify-write round: catalog + context -> validated overlay -> send. */
+    private fun applySetting(parameter: String, value: String): Either<ActuatorOperationFailure, Unit> =
+        runBlocking {
+            either {
+                val ref = ensureNotNull(appliances.refOf(deviceProviderId)) { AcApplianceNotSynced }
+                val catalog = client.loadCommands(ref).mapLeft(::HonAcProviderFailure).bind()
```
AGRFesta: 🤖 **Reviewer** — _severity: low · confidence: low_

**Observation (no change requested): per-write catalog re-fetch.** Every write costs `GET /retrieve` + `GET /context` + `POST /send`. The context read must stay fresh (it's the RMW base), but the catalog is firmware-static per appliance — a future "apply scene" flow (mode + temperature + fan) would hit the cloud 9 times. Fine for the current single-setting use; flagging it as a candidate follow-up (cache the catalog per appliance ref, invalidated on fwVersion change) rather than something to fix in this PR.

AGRFesta:
🤖 **Reviewer** — review summary (round 1, sha `b59cf540681121cad73209bcce12b8530d77c13d`)

Round 1 review of the hOn AC write path (#251). Solid work overall: the read-modify-write design is well documented, the wire protocol is pinned by slice tests over real captured fixtures, wind-direction sanitization and the fixed-parameter quirk are both guarded, and the TDD trail (declared green-on-arrival guards, batch markers) is visible in the tests. Findings: 2 medium (raw BigDecimal instead of the Temperature value object on the new core port; unguarded catalog-shape assumptions in AcSettingsCommand that can throw through the outbound port) and 5 low (changelog/docs coverage of the now-visible ACTUATOR feature and the hOn diagnostics example, mid-class companion object, duplicated context-query encoding, per-write catalog re-fetch).

CHANGELOG.md
AGRFesta: 🤖 **Author** — accepted, fixed in `17cb42c`.

Right — the #252 deferral covers the **control endpoints**, but the roles→features projection is observable today in `GET /devices`. Added the `Added` entry for the `ACTUATOR` feature.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/ports/outbounds/devices/AirConditioner.kt
AGRFesta: 🤖 **Author** — accepted, fixed in `17cb42c`.

The port now takes `Temperature`; the driver serializes `temperature.value.toPlainString()`, dropping the hand-rolled `stripTrailingZeros()` — the VO's `setScale(2, HALF_UP).stripTrailingZeros()` subsumes it and normalizes sub-cent noise at the domain boundary instead of letting it die as an opaque `AcSettingOutOfRange`. `HonAcTest` pins the wire form (`"23"`, never `"23.0"`) via `Temperature.of("23.0")`.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/devices/AcSettingsCommand.kt
AGRFesta: 🤖 **Author** — accepted, fixed in `17cb42c`, with a deliberate **lenient** semantic.

A malformed rule is *unenforceable*, so the value is admitted and the cloud stays the ultimate validator (a bad value comes back as `HonCommandRejected`, still a typed failure) — rejecting on drift would brick the control on exactly the payloads we cannot predict. Concretely:

* unreadable/missing range bounds → rule skipped (change admitted);
* `fixed` without `fixedValue` → `"0"` (addhOn's base-parameter default);
* empty `enumValues` → no rule at all — neither validation nor sanitization (this also removes the `allowed.first()` `NoSuchElementException` edge Gemini flagged).

Three new unit tests in `AcSettingsCommandTest` pin the drift cases (truncated range spec, fixed without value, empty enum on both a validated and a sanitized parameter).

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/devices/HonAc.kt
AGRFesta: 🤖 **Author** — accepted, fixed in `17cb42c`.

The provider-specific examples in `docs/api/devices.md` §diagnostics now name hOn (raw command context) alongside SwitchBot and Netatmo.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/devices/HonAc.kt
AGRFesta: 🤖 **Author** — accepted, fixed in `17cb42c`.

Added `rawContext(appliance: HonApplianceRef)` on the client, sharing a private `contextParams(...)` with `loadAttributes` so the wire encoding lives in one place; the map-based variant stays for the provider-inspector probes (arbitrary passthrough params). `HonAc.inspect()` now uses the overload, and the driver test still pins the query params.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/devices/HonAc.kt
AGRFesta: 🤖 **Author** — agreed, and deliberately deferred. The catalog is firmware-static per appliance, so a per-ref cache invalidated on `fwVersion` change is the right shape — but caching semantics (lifetime, invalidation on sync, memory bounds) deserve their own test list rather than riding this PR. Tracked as a follow-up candidate; today's cost is 3 calls per single-setting write, acceptable for the current single-setting use.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/devices/HonAc.kt
AGRFesta: 🤖 **Author** — accepted, fixed in `17cb42c`. Correct diagnosis: an un-tidied TDD append (the cycle that introduced `applySetting` planted the companion before the later setter cycles ran). The `AirConditioner` overrides are now grouped together, private helpers follow, companion at the bottom.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/ports/outbounds/devices/AirConditioner.kt
AGRFesta: 🤖 **Reviewer**

Verified: `AirConditioner.setTargetTemperature` now takes `Temperature`, and the driver serializes `temperature.value.toPlainString()`. Checked the VO: `Temperature.of` applies `setScale(2, HALF_UP).stripTrailingZeros()`, which both subsumes the deleted hand-rolled `stripTrailingZeros()` and normalizes unscaled input at the domain boundary. `HonAcTest` pins the wire form via `Temperature.of("23.0")` → `"23"`. Resolved.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/devices/AcSettingsCommand.kt
AGRFesta: 🤖 **Reviewer**

Verified: all three throw paths are gone — unreadable range bounds skip the rule, `fixed` without `fixedValue` falls back to `"0"`, empty `enumValues` is treated as no-rule (which also removes the `allowed.first()` NoSuchElementException edge). The lenient semantic is a sound call for this boundary: a malformed rule is unenforceable, the cloud remains the authoritative validator, and a bad value still surfaces as a typed failure rather than a crash — and the rationale is documented in the KDoc, not just the PR thread. Three drift tests pin the behaviour. Resolved.

CHANGELOG.md
AGRFesta: 🤖 **Reviewer**

Verified: the `Added` entry for the `ACTUATOR` feature is in `CHANGELOG.md` under `[Unreleased]`. Resolved.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/devices/HonAc.kt
AGRFesta: 🤖 **Reviewer**

Verified: the provider-specific examples in `docs/api/devices.md` §diagnostics now name hOn (raw command context). Resolved.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/devices/HonAc.kt
AGRFesta: 🤖 **Reviewer**

Verified: `rawContext(appliance: HonApplianceRef)` shares `contextParams(...)` with `loadAttributes`, so the wire encoding has a single source; keeping the map-based overload for the inspector's arbitrary passthrough probes is the right split. `HonAc.inspect()` uses the typed overload and the driver test still pins the query. Resolved.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/devices/HonAc.kt
AGRFesta: 🤖 **Reviewer**

Agreed — deferring the per-appliance catalog cache to its own issue (with its own test list for lifetime/invalidation semantics) is the right call; 3 calls per single-setting write is acceptable today. Closing as a tracked follow-up candidate. Resolved.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/devices/HonAc.kt
AGRFesta: 🤖 **Reviewer**

Verified: `AirConditioner` overrides are grouped, private helpers follow, companion sits at the bottom. Resolved.

AGRFesta:
🤖 **Reviewer** — review summary (round 2, sha `17cb42c181b94a3eb41823c632b347e40e6645c9`)

Round 2 re-review of the fixup delta (`b59cf54..17cb42c`). All six round-1 threads are addressed and verified: the port now takes the `Temperature` value object (verified: `Temperature.of` applies `setScale(2, HALF_UP).stripTrailingZeros()`, so the pinned wire form `"23"` holds and the hand-rolled normalization is gone); the catalog-drift reads are defensive with a deliberate, documented lenient semantic and three pinning tests (which also structurally remove the `allowed.first()` NoSuchElementException edge Gemini flagged); changelog and `docs/api/devices.md` cover the ACTUATOR feature and the hOn diagnostics example; the context-query encoding lives in one place (`contextParams` shared by `loadAttributes`/`rawContext`); the class layout is tidied; the per-write catalog re-fetch is agreed as a tracked follow-up. Both touched test classes pass locally. No new findings in the delta — review approved from my side; the three open Gemini threads are subsumed by the drift fix (two structurally, one deliberately resolved with lenient-admit instead of reject) and can be closed by the author.

Issue #246, opened 2026-07-16
[Epic] Haier hOn provider integration (AC monitoring & control)


AGRFesta:
## Goal

Add **hOn (Haier)** as a third device provider alongside SwitchBot and Netatmo, so that
appliances paired to the hOn account are discovered by the device sync and — for the air
conditioner — can be monitored (readings) and controlled (power, mode, target temperature,
fan speed) through our API.

## Background

The hOn cloud has **no official public API**. The integration protocol comes from the
open-source Home Assistant integration **addhOn** (https://github.com/tis24dev/addhOn, MIT,
by tis24dev), reverse-engineered from the hOn mobile app. That repo also ships
`examples/ktor-client/` — a dependency-free Kotlin/Ktor port of the transport layer
(Salesforce OAuth login + authenticated API calls) written to be embedded in a Spring Boot
app, with a README documenting every server quirk. We will **translate and adapt** that
code (kotlinx-serialization → Jackson, our failure model, our conventions) rather than
depend on it.

The `addhOn/` folder currently in the repo root is a **temporary reference checkout** — it
must never be committed; the canonical reference is the upstream repo.

### Protocol in a nutshell

1. **Auth, level 1 — Salesforce OAuth** on `account2.hon-smarthome.com`: not a clean OAuth
   flow but the mobile app's web login reproduced by scraping the Salesforce login pages
   (regex extraction, session cookies, byte-exact `urllib.quote`-style percent-encoding).
   Produces `access_token`, `refresh_token`, `id_token`.
2. **Auth, level 2 — IoT login**: `POST https://api-iot.he.services/auth/v1/login` with the
   `id-token` header exchanges it for a **cognito-token**.
3. Every API call carries **both** `cognito-token` and `id-token` headers.
4. Tokens last ~8h (renew at 7h). The **refresh token rotates**: the rotated value must be
   persisted or the next refresh fails.
5. Device state is polled via `GET /commands/v1/context`; commands go through
   `POST /commands/v1/send` and must be validated against the per-device command catalog
   (`GET /commands/v1/retrieve`).

## Phase 0 — prerequisites (human, blocking)

- [x] **Account access**: the account was created via Google social login, which the
  programmatic flow does **not** support (it posts email+password to the Salesforce form).
  Resolved via the hOn "forgot password" flow on the existing account (pairing kept).
  ⚠️ Follow-up before #247: rotate the password (the smoke-test one was exposed during
  debugging) and put the final value in the server configuration only.
[...]

PR #260, opened 2026-07-19
issue #252: hOn AC control REST endpoints — control-state read, singl…

### LINKED ISSUE #252: AC control REST endpoints + API docs + Bruno

AGRFesta:
…e-command partial update

### DIFF
--- a/app/src/main/kotlin/org/agrfesta/sh/api/controllers/AirConditionerController.kt
+++ b/app/src/main/kotlin/org/agrfesta/sh/api/controllers/AirConditionerController.kt
@@ -1,0 +1,120 @@
+package org.agrfesta.sh.api.controllers
+
+import arrow.core.Either
+import org.agrfesta.sh.api.core.application.ports.inbounds.devices.GetAcStateUseCase
+import org.agrfesta.sh.api.core.application.ports.inbounds.devices.SetAcSettingsUseCase
+import org.agrfesta.sh.api.core.domain.commons.Temperature
+import org.agrfesta.sh.api.core.domain.devices.AcFanSpeed
+import org.agrfesta.sh.api.core.domain.devices.AcMode
+import org.agrfesta.sh.api.core.domain.devices.AcPowerCommand
+import org.agrfesta.sh.api.core.domain.devices.AcSettingsUpdate
+import org.agrfesta.sh.api.core.domain.devices.AcState
+import org.agrfesta.sh.api.core.domain.devices.ActuatorStatus
+import org.agrfesta.sh.api.core.domain.failures.AcProviderFailure
+import org.agrfesta.sh.api.core.domain.failures.DeviceNotFound
+import org.agrfesta.sh.api.core.domain.failures.DeviceRepositoryError
+import org.agrfesta.sh.api.core.domain.failures.InvalidAcSetting
+import org.agrfesta.sh.api.core.domain.failures.NotAnAirConditioner
+import org.springframework.http.HttpStatus
+import org.springframework.http.ResponseEntity
+import org.springframework.http.ResponseEntity.internalServerError
+import org.springframework.http.ResponseEntity.noContent
+import org.springframework.http.ResponseEntity.notFound
+import org.springframework.http.ResponseEntity.ok
+import org.springframework.http.ResponseEntity.status
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.PatchMapping
+import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+import java.math.BigDecimal
+import java.util.UUID
+
+@RestController
+@RequestMapping("/devices")
+class AirConditionerController(
+    private val getAcStateUseCase: GetAcStateUseCase,
+    private val setAcSettingsUseCase: SetAcSettingsUseCase
+) {
+    @PatchMapping("/{uuid}/air-conditioner")
+    fun setAcSettings(
+        @PathVariable uuid: UUID,
+        @RequestBody request: AcSettingsRequest
+    ): ResponseEntity<Any> {
+        val update = try {
+            request.toUpdate()
+        } catch (e: IllegalArgumentException) {
+            return status(HttpStatus.BAD_REQUEST).body(MessageResponse(e.message ?: "Invalid request body"))
+        }
+        if (update == AcSettingsUpdate()) {
+            return status(HttpStatus.BAD_REQUEST).body(MessageResponse("At least one field must be provided"))
+        }
+        return when (val result = setAcSettingsUseCase.execute(uuid, update)) {
+            is Either.Right -> noContent().build()
+            is Either.Left -> when (val failure = result.value) {
+                is DeviceNotFound -> notFound().build()
+                DeviceRepositoryError -> internalServerError()
+                    .body(MessageResponse("Unable to retrieve device '$uuid'!"))
+                NotAnAirConditioner -> conflictNotAnAc(uuid)
+                is InvalidAcSetting -> status(HttpStatus.BAD_REQUEST)
+                    .body(MessageResponse(failure.reason))
+                is AcProviderFailure -> badGateway(failure.message, uuid)
+            }
+        }
+    }
+
+    @GetMapping("/{uuid}/air-conditioner")
+    fun getAcState(@PathVariable uuid: UUID): ResponseEntity<Any> =
+        when (val result = getAcStateUseCase.execute(uuid)) {
+            is Either.Right -> ok(result.value.toResponse())
+            is Either.Left -> when (val failure = result.value) {
+                is DeviceNotFound -> notFound().build()
+                DeviceRepositoryError -> internalServerError()
+                    .body(MessageResponse("Unable to retrieve device '$uuid'!"))
+                NotAnAirConditioner -> conflictNotAnAc(uuid)
+                is AcProviderFailure -> badGateway(failure.message, uuid)
+            }
+        }
+
+    private fun conflictNotAnAc(uuid: UUID): ResponseEntity<Any> =
+        status(HttpStatus.CONFLICT).body(MessageResponse("Device '$uuid' is not an air conditioner!"))
+
+    private fun badGateway(message: String?, uuid: UUID): ResponseEntity<Any> =
+        status(HttpStatus.BAD_GATEWAY)
+            .body(MessageResponse(message ?: "Air conditioner provider failed for device '$uuid'!"))
+}
+
+private fun AcSettingsRequest.toUpdate() = AcSettingsUpdate(
+    power = power?.let { parseEnum<AcPowerCommand>("power", it) },
+    mode = mode?.let { parseEnum<AcMode>("mode", it) },
+    targetTemperature = targetTemperature?.let { Temperature.of(it) },
+    fanSpeed = fanSpeed?.let { parseEnum<AcFanSpeed>("fanSpeed", it) },
+)
+
+private inline fun <reified E : Enum<E>> parseEnum(field: String, value: String): E =
+    enumValues<E>().firstOrNull { it.name == value }
+        ?: throw IllegalArgumentException(
+            "Invalid $field '$value', allowed: ${enumValues<E>().joinToString()}",
+        )
+
+data class AcSettingsRequest(
+    val power: String? = null,
+    val mode: String? = null,
+    val targetTemperature: BigDecimal? = null,
+    val fanSpeed: String? = null
+)
+
+data class AcStateResponse(
+    val power: ActuatorStatus,
+    val mode: AcMode?,
+    val targetTemperature: BigDecimal?,
+    val fanSpeed: AcFanSpeed?
+)
+
+fun AcState.toResponse() = AcStateResponse(
+    power = power,
+    mode = mode,
+    targetTemperature = targetTemperature?.value,
+    fanSpeed = fanSpeed
+)
--- a/app/src/test/kotlin/org/agrfesta/sh/api/DevicesIntegrationTest.kt
+++ b/app/src/test/kotlin/org/agrfesta/sh/api/DevicesIntegrationTest.kt
@@ -1,17 +1,22 @@
 package org.agrfesta.sh.api
 
 import arrow.core.getOrElse
+import arrow.core.right
 import com.fasterxml.jackson.databind.JsonNode
 import com.fasterxml.jackson.databind.ObjectMapper
 import io.kotest.assertions.withClue
 import io.kotest.matchers.collections.shouldContainExactly
 import io.kotest.matchers.collections.shouldContainExactlyInAnyOrder
 import io.kotest.matchers.collections.shouldHaveSize
+import io.kotest.matchers.comparables.shouldBeEqualComparingTo
+import io.kotest.matchers.nulls.shouldNotBeNull
 import io.kotest.matchers.shouldBe
 import io.mockk.coEvery
+import io.mockk.coVerify
 import io.mockk.every
 import io.restassured.RestAssured.given
 import io.restassured.http.ContentType
+import org.agrfesta.sh.api.controllers.AcStateResponse
 import org.agrfesta.sh.api.controllers.AssignmentResponse
 import org.agrfesta.sh.api.controllers.DeviceResponse
 import org.agrfesta.sh.api.controllers.DeviceViewResponse
@@ -27,9 +32,13 @@ import org.agrfesta.sh.api.core.application.ports.outbounds.devices.DevicesRepos
 import org.agrfesta.sh.api.core.application.readmodels.devices.AssignmentRole
 import org.agrfesta.sh.api.core.domain.alerts.AlertTarget
 import org.agrfesta.sh.api.core.domain.alerts.AlertType
+import org.agrfesta.sh.api.core.domain.devices.AcFanSpeed
+import org.agrfesta.sh.api.core.domain.devices.AcMode
+import org.agrfesta.sh.api.core.domain.devices.ActuatorStatus
 import org.agrfesta.sh.api.core.domain.devices.DeviceFeature.SENSOR
 import org.agrfesta.sh.api.core.domain.devices.DeviceModel
 import org.agrfesta.sh.api.core.domain.devices.DeviceStatus
+import org.agrfesta.sh.api.core.domain.devices.Provider.HON
 import org.agrfesta.sh.api.core.domain.devices.Provider.NETATMO
 import org.agrfesta.sh.api.core.domain.devices.Provider.SWITCHBOT
 import org.agrfesta.sh.api.core.domain.devices.ProviderDeviceData
@@ -39,6 +48,9 @@ import org.agrfesta.sh.api.domain.aSensorProviderData
 import org.agrfesta.sh.api.domain.anAlert
 import org.agrfesta.sh.api.domain.anArea
 import org.agrfesta.sh.api.persistence.jdbc.repositories.DevicesJdbcRepository
+import org.agrfesta.sh.api.providers.hon.HonApplianceRef
+import org.agrfesta.sh.api.providers.hon.HonApplianceStore
+import org.agrfesta.sh.api.providers.hon.HonService
 import org.agrfesta.sh.api.providers.netatmo.NetatmoIntegrationAsserter
 import org.agrfesta.sh.api.providers.switchbot.SwitchBotDeviceType
 import org.agrfesta.sh.api.providers.switchbot.aSwitchBotDevice
@@ -46,6 +58,7 @@ import org.agrfesta.sh.api.providers.switchbot.aSwitchBotDevicesListSuccessRespo
 import org.agrfesta.test.mothers.aRandomUniqueString
 import org.junit.jupiter.api.BeforeEach
 import org.junit.jupiter.api.Test
+import java.math.BigDecimal
 import java.time.Instant
 import java.time.temporal.ChronoUnit
 import java.util.UUID
@@ -58,9 +71,11 @@ class DevicesIntegrationTest(
     private val alertsRepository: AlertsRepository,
     private val objectMapper: ObjectMapper,
     private val netatmoIntegrationAsserter: NetatmoIntegrationAsserter,
-    private val catalog: DeviceModelCatalog
+    private val catalog: DeviceModelCatalog,
+    private val honApplianceStore: HonApplianceStore
 ) : AbstractIntegrationTest() {
     private val now = Instant.now()
+    private val honAcModel = DeviceModel(HonService.AC_AS35PBPHRA_PRE_MODEL)
 
     @BeforeEach
     fun init() {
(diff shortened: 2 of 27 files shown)


core/src/main/kotlin/org/agrfesta/sh/api/core/domain/devices/AcSettingsUpdate.kt
```diff
@@ -0,0 +1,16 @@
...
+import org.agrfesta.sh.api.core.domain.commons.Temperature
+
+/**
+ * Partial update of an air conditioner's settings: only non-null fields are commands, the
+ * device keeps its current value for everything else (the driver's read-modify-write fills
+ * the rest). [power] is a command, so only [ActuatorStatus.ON] and [ActuatorStatus.OFF] are
+ * meaningful values for it.
+ */
+data class AcSettingsUpdate(
+    val power: ActuatorStatus? = null,
```
AGRFesta: 🤖 **Reviewer** — _severity: medium · confidence: high_

❗️ **Exception can cross the port boundary.** `power: ActuatorStatus?` admits `UNDEFINED`, which the KDoc forbids but the type allows. The hOn driver enforces it with `error("UNDEFINED is not a power command")` (`HonAcCodes.kt`), so `AirConditioner.updateSettings(AcSettingsUpdate(power = UNDEFINED))` throws an `IllegalStateException` out of an outbound port — violating the architecture rule that ports "must not throw technical exceptions; use Either" (ARCHITECTURE.md §2). Today the only caller is the controller, which rejects `UNDEFINED` at parse time, but any future core caller (e.g. a scheduler driving the AC) hits the crash path.

Suggestion: make the illegal state unrepresentable — a dedicated command type (`enum class AcPowerCommand { ON, OFF }` or simply `Boolean`) instead of reusing the reading enum. That deletes the `error()` branch, the KDoc caveat, and the controller's special-cased `parsePower` in one move.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/devices/GetAcStateService.kt
```diff
@@ -0,0 +1,40 @@
...
+                    is DeviceNotFound -> failure
+                    DeviceRepositoryError -> DeviceRepositoryError
+                }
+            }
+            .flatMap { device ->
+                val factory = deviceFactories.firstOrNull { it.provider == device.provider }
+                    ?: return@flatMap NotAnAirConditioner.left()
+                val driver = factory.createDevice(device) as? AirConditioner
+                    ?: return@flatMap NotAnAirConditioner.left()
+                driver.getState().mapLeft { AcProviderFailure(it.toString()) }
```
AGRFesta: 🤖 **Reviewer** — _severity: medium · confidence: high_

**Debug `toString()` as a client-facing message.** `AcProviderFailure(it.toString())` puts the Kotlin data-class debug representation on the wire: a 502 body of `{"message": "HonAcProviderFailure(failure=HonNetworkError(reason=Connection reset))"}` or `{"message": "AcApplianceNotSynced"}`. Same in `SetAcSettingsService` (`else -> AcProviderFailure(failure.toString())`).

The sibling path from #251 set a better precedent: `InspectDeviceService` maps `DevicesProviderError` to `DiagnosticsProviderFailure(failure.exception.message)` — a human message, not a debug dump. The docs promise `"<provider failure>"`, so nothing is broken, but internal type names are now API-observable (clients will match on them) and nested `reason` strings carry raw exception text. Consider giving `ActuatorOperationFailure` a `message`-style member (as `AcSettingRejected.reason` already does for the rejection case), or mapping the known cases explicitly.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/devices/GetAcStateService.kt
```diff
@@ -0,0 +1,40 @@
...
+    override fun execute(deviceId: UUID): Either<GetAcStateFailure, AcState> =
+        devicesRepository.getDeviceById(deviceId)
+            .mapLeft { failure ->
+                when (failure) {
+                    is DeviceNotFound -> failure
+                    DeviceRepositoryError -> DeviceRepositoryError
+                }
+            }
+            .flatMap { device ->
+                val factory = deviceFactories.firstOrNull { it.provider == device.provider }
```
AGRFesta, line 34: 🤖 **Reviewer** — _severity: medium · confidence: high_

**Third copy of the resolve-device→find-factory→create-driver block.** `InspectDeviceService`, `GetAcStateService` and `SetAcSettingsService` now repeat the same sequence: `getDeviceById` + identical repository-failure `mapLeft` + `deviceFactories.firstOrNull { it.provider == device.provider }` + `createDevice` + type check. Rule of three reached — Phase 3 (Refactor) is supposed to remove cross-class duplication.

Suggestion: extract an application-layer collaborator (e.g. a `DeviceDriverResolver` returning `Either<ResolutionFailure, D>` with a reified/`KClass` driver-type parameter), and let each use case map `ResolutionFailure` into its own sealed contract. Fine as a follow-up issue if you'd rather not grow this PR, but it should not reach a fourth copy.

app/src/main/kotlin/org/agrfesta/sh/api/controllers/AirConditionerController.kt
```diff
@@ -0,0 +1,126 @@
...
+        }
+        return when (val result = setAcSettingsUseCase.execute(uuid, update)) {
+            is Either.Right -> noContent().build()
+            is Either.Left -> when (val failure = result.value) {
+                is DeviceNotFound -> notFound().build()
+                DeviceRepositoryError -> internalServerError()
+                    .body(MessageResponse("Unable to retrieve device '$uuid'!"))
+                NotAnAirConditioner -> conflictNotAnAc(uuid)
+                is InvalidAcSetting -> status(HttpStatus.BAD_REQUEST)
+                    .body(MessageResponse(failure.reason))
```
AGRFesta, line 61: 🤖 **Reviewer** — _severity: low · confidence: medium_

**409 vs 501 inconsistency for "provider has no driver factory".** `NotAnAirConditioner` folds two situations into one 409: (a) the device is genuinely not an AC — a client error, 409 is defensible; (b) no `ProviderDevicesFactory` is registered for the device's provider — a server capability gap, which the sibling endpoint on the same resource (`GET /devices/{uuid}/diagnostics`) maps to **501 NOT_IMPLEMENTED** (`DiagnosticsNotSupported`). The same underlying condition now yields 409 on one sub-resource and 501 on another. It is documented in `devices.md`, so it's a deliberate contract — but if the distinction was intentional, a line in the docs saying why would help; otherwise consider splitting the no-factory case to match the diagnostics precedent.

app/src/main/kotlin/org/agrfesta/sh/api/controllers/AirConditionerController.kt
```diff
@@ -0,0 +1,126 @@
...
+
+@RestController
+@RequestMapping("/devices")
+class AirConditionerController(
+    private val getAcStateUseCase: GetAcStateUseCase,
+    private val setAcSettingsUseCase: SetAcSettingsUseCase
+) {
+    @PatchMapping("/{uuid}/air-conditioner")
+    fun setAcSettings(
+        @PathVariable uuid: UUID,
```
AGRFesta, line 42: 🤖 **Reviewer** — _severity: medium · confidence: medium_

**No integration test for the two new flows.** The TDD infrastructure table (TDD_WORKFLOW.md) prescribes one integration test per flow covering the happy path, verifying wiring between components — and #251 honored it for diagnostics (`DevicesIntegrationTest."GET device diagnostics returns the provider raw payload verbatim"`). The AC endpoints have unit + MVC-slice coverage only, so what's left unverified end-to-end is exactly the wiring: `Set<ProviderDevicesFactory>` injection into the two new services and the hOn factory producing an `AirConditioner` driver. A GET happy-path and a PATCH happy-path in `DevicesIntegrationTest` (reusing the hOn mock setup from the diagnostics test) would close the gap.

AGRFesta:
🤖 **Reviewer** — review summary (round 1, sha `d0d8c64344de94c1acdf57c1703fc4c8a44c7412`)

Round 1 review of the AC control REST endpoints (#252). Solid PR overall: ports/use-cases follow the conventions (execute(), one service per interface, sealed use-case failure contracts, non-sealed port failure with rationale), docs/Bruno/changelog are all updated, and the provider tests pin the wire contract against real captured payloads. Findings are design/consistency level: a port contract that can throw across the boundary, a debug toString() surfaced as an API message, a third copy of the device→factory→driver resolution block, a 409-vs-501 inconsistency with the sibling diagnostics endpoint, and missing integration coverage for the two new flows.

core/src/main/kotlin/org/agrfesta/sh/api/core/domain/devices/AcSettingsUpdate.kt
AGRFesta: 🛠️ **Author** — accepted, fixed in `889eba1`.

Reso l'illegal state irrappresentabile: nuovo `enum class AcPowerCommand { ON, OFF }` al posto di `ActuatorStatus?` in `AcSettingsUpdate`. Spariscono la branch `error("UNDEFINED is not a power command")` nel driver, il caveat KDoc e il `parsePower` speciale del controller (ora un normale `parseEnum<AcPowerCommand>`). Nessun percorso può più far attraversare un'eccezione al port.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/devices/GetAcStateService.kt
AGRFesta: 🛠️ **Author** — accepted, fixed in `889eba1`.

`ActuatorOperationFailure` ha ora `val message: String?` (default `null`), stesso pattern di `AcSettingRejected.reason`. Gli adapter hOn lo riempiono col messaggio umano già esistente (`toException().message`); `GetAcStateService`/`SetAcSettingsService` mappano `it.message` invece del debug `toString()`. Effetto collaterale corretto: `NetatmoContractBreak.message` ora ha `override`.

app/src/main/kotlin/org/agrfesta/sh/api/controllers/AirConditionerController.kt
AGRFesta: 🛠️ **Author** — accepted, fixed in `889eba1`.

La scelta 409 è deliberata: ho aggiunto in entrambe le sezioni `409` di `devices.md` la riga che la motiva — per il client il fatto azionabile è "questo device non è pilotabile come AC", distinto dal capability-gap diagnostico che giustifica il 501 su `GET /devices/{uuid}/diagnostics`.

app/src/main/kotlin/org/agrfesta/sh/api/controllers/AirConditionerController.kt
AGRFesta: 🛠️ **Author** — accepted, fixed in `889eba1`.

Aggiunti due happy-path end-to-end in `DevicesIntegrationTest` (`GET` + `PATCH /devices/{uuid}/air-conditioner`) che esercitano il vero `HonDevicesFactory`→`HonAc`. Per farlo hOn è ora abilitato nel profilo test con `HonApiClient` mockato in `AbstractIntegrationTest` (context singolo preservato, come per SwitchBot); un default `loadAppliances()→[]` lascia `synchronize()` e gli altri test invariati.

Nota: il test diagnostics citato usa SwitchBot, non hOn — non c'era un setup hOn da riusare, l'ho introdotto. Il wire contract completo (38 parametri, sanitizzazione, normalizzazioni) resta pinnato su `HonAcTest`.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/devices/GetAcStateService.kt
AGRFesta: 🛠️ **Author** — accepted, tracciato come follow-up (non in questa PR).

Rule of three raggiunta. Estrarre un `DeviceDriverResolver` (resolve device → find factory → create+typecheck driver → `Either<ResolutionFailure, D>` con tipo reificato) tocca però anche `InspectDeviceService` (già in main da #251), quindi lo tengo fuori da questa PR per non allargarne lo scope. Apro una issue di follow-up dedicata, prima che arrivi una quarta copia.

core/src/main/kotlin/org/agrfesta/sh/api/core/domain/failures/ActuatorOperationFailure.kt
```diff
@@ -6,4 +6,22 @@ package org.agrfesta.sh.api.core.domain.failures
...
  * with provider-specific subtypes; callers handle it generically.
  */
-interface ActuatorOperationFailure
+interface ActuatorOperationFailure {
+    /**
+     * Optional human-readable cause, suitable to surface to API clients. Null when the
+     * adapter has nothing better than its type name — callers fall back to a generic text,
+     * never to the debug `toString()`.
+     */
+    val message: String? get() = null
```
AGRFesta, line 15: 🤖 **Reviewer** — _severity: low · confidence: medium_

**`message` and `reason` now overlap on the same hierarchy.** `AcSettingRejected.reason: String` and the new `ActuatorOperationFailure.message: String?` express the same thing — a human-readable cause suitable for API clients — but they are independent members, and none of the three hOn rejection subtypes (`AcSettingNotSupported`, `AcSettingOutOfRange`, `AcUnknownParameter`) overrides `message`. So a rejection carries a perfectly good `reason` while its `message` is `null`. It works today only because `SetAcSettingsService` special-cases `is AcSettingRejected` before the `else` branch; any other caller reading the generic member (the GET path already does, and a future scheduler driving the AC would too) silently degrades to the generic "Air conditioner provider failed" text and loses the reason.

One line ties them and removes the trap: `interface AcSettingRejected : ActuatorOperationFailure { val reason: String; override val message: String get() = reason }`.

app/src/test/kotlin/org/agrfesta/sh/api/DevicesIntegrationTest.kt
```diff
@@ -231,6 +246,85 @@ class DevicesIntegrationTest(
...
     }
 
+    @Test
+    fun `GET air-conditioner returns the hOn driver control state end-to-end`() {
+        val mac = aRandomUniqueString()
+        val deviceId = UUID.randomUUID()
+        devicesDao.create(deviceId, aProviderDeviceData(provider = HON, providerId = mac, model = honAcModel))
+            .getOrElse { error("Failed to create device: $it") }
+        // The appliance-list fields the driver needs are populated by the device sync at runtime.
+        honApplianceStore.save(HonApplianceRef(macAddress = mac, applianceType = "AC"))
```
AGRFesta, line 256: 🤖 **Reviewer** — _severity: low · confidence: medium_

**`HonApplianceStore` is shared mutable state across the cached context and is never reset.** It is a singleton `@Service` backed by a `ConcurrentHashMap`, so entries saved here outlive the test and accumulate for every test sharing the context — unlike the DB (`devicesRepository.deleteAll()`) and Redis (`flushDb()`), which `@BeforeEach` does clean. No test can fail today because the MAC comes from `aRandomUniqueString()`, so a stale ref is unreachable; the risk is a later test that pins a fixed MAC and inherits a ref it never saved. Cheapest fix is a `clear()` on the store called from the same `@BeforeEach` that already flushes Redis.

providers/src/main/kotlin/org/agrfesta/sh/api/providers/hon/devices/HonAcFailure.kt
```diff
@@ -11,22 +13,33 @@ data class AcSettingNotSupported(
...
+    override val message: String? get() = failure.toException().message
+}
 
 /**
  * The appliance is not in the [org.agrfesta.sh.api.providers.hon.HonApplianceStore] yet: the
  * store is in-memory and empty after a restart until the device sync runs.
  */
-data object AcApplianceNotSynced : HonAcFailure
+data object AcApplianceNotSynced : HonAcFailure {
+    override val message: String get() = "Appliance not synced yet: run the device synchronization first"
```
AGRFesta, line 44: 🤖 **Reviewer** — _severity: low · confidence: low_

**This message is operator-facing, but it reaches an API client.** `"Appliance not synced yet: run the device synchronization first"` now travels into a `502` body via `AcProviderFailure`. "Run the device synchronization" is an instruction the HTTP client cannot act on — it describes an internal server lifecycle (in-memory store empty until the sync scheduler has run). Something like "Device not available yet, retry shortly" tells the client the one thing it can do, while the operational detail stays where it belongs, in the logs and in the KDoc above. Very much a judgement call — flagging it because this is the first `ActuatorOperationFailure.message` that is written for a human rather than derived from a transport error.

core/src/main/kotlin/org/agrfesta/sh/api/core/domain/devices/AcSettingsUpdate.kt
AGRFesta: 🤖 **Reviewer**

Verified in `889eba1`. `AcPowerCommand { ON, OFF }` replaces `ActuatorStatus?` in `AcSettingsUpdate`; `HonAcCodes.honCode()` is now a total two-branch `when` with no `error()`, so no path can throw across the outbound port. The controller's `parsePower` is gone and `power` goes through the same `parseEnum` as the other fields — and the client-visible 400 text is byte-identical to before (`Invalid power 'X', allowed: ON, OFF`), so the documented contract did not move. `HonAc.on()`/`off()` and the tests are consistent. Exactly the "make the illegal state unrepresentable" shape the comment asked for.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/devices/GetAcStateService.kt
AGRFesta: 🤖 **Reviewer**

Verified in `889eba1`. `ActuatorOperationFailure.message: String?` (default `null`) is populated by both hOn subtypes, `GetAcStateService`/`SetAcSettingsService` map `it.message`, and `AirConditionerController.badGateway` falls back to a generic text when it is `null` — so a debug `toString()` can no longer reach the wire, and neither can a `null` message. I also checked `HonFailure.toException()` for what is now client-observable: every branch is a hand-written sentence, no response bodies or token context leak into the 502. The `override` on `NetatmoContractBreak.message` is the correct consequence.

One loose end from this change, raised separately on `ActuatorOperationFailure.kt`: the rejection subtypes carry `reason` but leave the new `message` null.

core/src/main/kotlin/org/agrfesta/sh/api/core/application/usecases/devices/GetAcStateService.kt
AGRFesta: 🤖 **Reviewer**

Agreed on deferring — the extraction reaches `InspectDeviceService`, which is already on `main` from #251, so it is genuinely outside this PR's diff and would widen the review surface for no behavioural gain.

Closing this thread on the understanding that the follow-up gets tracked: `gh issue list` shows no open issue for the `DeviceDriverResolver` extraction yet (only #252 and the #246 epic), so the note lives only in this thread for now. Worth opening before the fourth copy shows up.

app/src/main/kotlin/org/agrfesta/sh/api/controllers/AirConditionerController.kt
AGRFesta: 🤖 **Reviewer**

Verified in `889eba1`. Both `409` sections of `devices.md` now state the rationale, and it is the right one: the client-actionable fact is "this device cannot be driven as an air conditioner", which is a different question from the diagnostics capability gap that justifies the `501`. That is what the comment was asking for — the inconsistency was only ever a problem if it was accidental.

app/src/main/kotlin/org/agrfesta/sh/api/controllers/AirConditionerController.kt
AGRFesta: 🤖 **Reviewer**

Verified in `889eba1`. The two new tests in `DevicesIntegrationTest` go through the real `HonDevicesFactory` -> `HonAc`, so the `Set<ProviderDevicesFactory>` injection and the driver creation the comment called out are now covered end-to-end; only the transport is mocked. The setup respects the context-cache convention (`HonApiClient` as a `@MockkBean` on `AbstractIntegrationTest`, not per-test), and the `loadAppliances() -> []` default is sufficient to leave the existing sync assertions untouched — `HonService` calls no other client method on that path, so no pre-existing test can hit an unstubbed call.

And you are right that the diagnostics precedent I cited was SwitchBot-based; introducing the hOn setup was the correct reading of the intent. One hygiene note on the shared `HonApplianceStore` is filed separately on the test file.

AGRFesta:
🤖 **Reviewer** — review summary (round 2, sha `889eba138e4cc1687767b258da7ffa9797f89ea8`)

Round 2 review, scoped to the delta d0d8c64..889eba1 plus the five open threads. All five round-1 findings are addressed: AcPowerCommand makes the UNDEFINED-as-command state unrepresentable and deletes both the driver's error() branch and the controller's special-cased parsePower (the client-visible 400 text is unchanged); ActuatorOperationFailure.message replaces the debug toString() on both service paths, with the controller falling back to a generic text when it is null; the 409-vs-501 rationale is documented in both sections of devices.md; and two end-to-end happy paths now exercise the real HonDevicesFactory -> HonAc wiring, with the hOn transport mocked in AbstractIntegrationTest so the Spring context stays cached. The resolver extraction is accepted as a follow-up. compileTestKotlin + detekt are green on this SHA. Three new low-severity findings on the delta, all non-blocking.

